PR #2, opened 2014-05-11
Fit map zoom and center to given bounds

sgillies:
Screenshot of code in action:

sheely

I'm using my Shapely library in this example to perform a "sloppy" 100 meter buffer on a GeoJSON polygon and then fit the map to the buffered extents.

I notice that until the map is rendered via `display(m)` the model's `_south`, `_west`, `_north`, and `_east` have only the default values. I'd like to be able to fit to bounds without having to display first, but I'm not sure how to do that yet.

### DIFF
--- a/leafletwidget/leaflet.py
+++ b/leafletwidget/leaflet.py
@@ -1,6 +1,7 @@
 from __future__ import print_function
 
 import json
+import math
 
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
Closing since this has not moved in 18 months.

PR #11, opened 2016-04-18
jupyter-js-widgets 5.0 migration

SylvainCorlay:
I spent some time today resurrecting the leaflet widget, and make it compatible with the next release of ipywidgets.

- The source is split into the javascript side, in the form of a proper npm package "jupyter-leaflet", and the python side "ipyleaflet". 

    The directory structure follows the same scheme as the beta versions of bqplot and pythreejs:
    ```javascript
    - js
        - package.json
        - webpack.config.js     // generates files in dist and in ipyleaflet/static
        - src/
            - extension.js
            - index.js
            - jupyter-leaflet.js
        - dist/                 // (generated)
            - contains the index.js to be used for static embedding of jupyter-leaflet
    - ipyleaflet
        - python...
    ```
- On the javascript side:

    - Like in bqplot and pythreejs, webpack produces 2 bundles for the notebook (extension and amd module), and one bundle for the embedding in static web pages.
    - The embedding of the widget in static webpages works. See example at: http://perso.crans.org/corlay/leaflet.html
    - The widget was probably not updated since IPython 3 because it used the old API before loading widget models became async because of the custom deserialization mechanism. I had to propagate that asynchronicity  a bit.
    - The view management for layers and controls is now based on the generic mechanism of ViewList (the old `dodiff` function that was used does not exist in ipywidgets since 4.x.)
   - All the widgets have custom models defining js-side default values to work with widget persistence.

- On the python side
    - we now have a proper `_version.py` mechanism.
    - `setup.py` successfully produces a wheel and an egg containing the built bundles for the notebook.
    -  it now uses the new traitlets API and does not rely on the deprecated `width` and `height` APIs.

- Finally we now use the new nbextension mechanism.
- We are not using the model and view registry anymore but `_[view/model]_[name/module]` mechanism.


SylvainCorlay:
cc @ellisonbg @jdfreder

PR #12, opened 2016-04-19
Showing all tiles on resize using leaflet's invalidateSize

SylvainCorlay:


SylvainCorlay:
This fixes a bug with certain tiles not being rendered.

PR #18, opened 2016-04-26
Use webpack to bundle the marker static resources.

SylvainCorlay:
@blink1073 this is what we talked about. I added some comments about the webpack issues in the webpack config file.

This is the version that generates the corrupted files.

### DIFF
--- a/js/webpack.config.js
+++ b/js/webpack.config.js
@@ -1,9 +1,30 @@
 var version = require('./package.json').version;
 
+var leaflet_marker_selector = /leaflet\/dist\/images\/marker-.*\.png/;
+
 var loaders = [
-    { test: /\.json$/, loader: "json-loader" },
-    { test: /\.css$/, loader: "style-loader!css-loader" },
-    { test: /\.(jpg|png|gif)$/, loader: "file" },
+    { test: /\.json$/, loader: 'json-loader' },
+    { test: /\.css$/, loader: 'style-loader!css-loader' },
+
+    // Generic file loader, Should be used for anything but leaflet's
+    // marker-icon.png, marker-icon-2x.png or marker-shadow.png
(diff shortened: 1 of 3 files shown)


js/webpack.config.js
```diff
@@ -3,7 +3,24 @@ var version = require('./package.json').version;
 var loaders = [
     { test: /\.json$/, loader: "json-loader" },
     { test: /\.css$/, loader: "style-loader!css-loader" },
+    // Generic file loader, Should be used for anything but leaflet's
+    // marker-icon.png, marker-icon-2x.png or marker-shadow.png
     { test: /\.(jpg|png|gif)$/, loader: "file" },
+    // Files marker-icon.png, marker-icon-2x.png or marker-shadow.png
```
blink1073: You can add an `exclude` option to the previous loader, no?

js/webpack.config.js
```diff
@@ -3,7 +3,24 @@ var version = require('./package.json').version;
...
+    // var path = L.Icon.Default.imagePath;
+    //
+    //  if (!path) {
+    //      throw new Error('Couldn\'t autodetect
+    //      L.Icon.Default.imagePath, set it manually.');
+    //  }
+    //
+    //  return path + '/marker-' + name + '.png';
+    //
+    { test: /\.(png)$/, loader: "file?name=[name].[ext]" }
```
blink1073: And then add `marker` to the regex here.

js/webpack.config.js
blink1073: http://webpack.github.io/docs/configuration.html#module-loaders

SylvainCorlay:
Thanks to @blink1073 I could get the marker-static assets to work!

blink1073:
:fist: :tada:

Issue #22, opened 2016-05-07
mouseover event not handled properly on some GeoJSON


galleon:
mouseover event not handled properly on some GeoJSON.

PR #23, opened 2016-05-07
fix LeafletGeoJSONView for mouseover event

galleon:
provide a fix that handle mouveover event (see #22) 
add an example using world countries


SylvainCorlay:
Thanks! Quick question: the example geojson is 56k lines long. I assume that it is ok. Although, is there an online resource from which we could request such files in the example notebook?

galleon:
I used this site https://geojson-maps.kyd.com.au/ to generate a world map at the lowest resolution. As stated: 
> These maps are sourced from Natural Earth and are in the public domain, so you are free to use them in any way without attribution.

A smaller one can be generated focusing on one continent if needed - or the example can simply be removed.

SylvainCorlay:
ok, I guess it would be better to have a smaller region - minimum resolution.

The PR looks good to me.

galleon:
@SylvainCorlay PR updated with Europe only - minimum resolution

SylvainCorlay:
👍 it just works! Merging.

FYI, since you opened this from the master branch, after this is merged, rather than pulling from master in the upstream repo, I recommend you do

```
git fetch upstream
git reset upstream/master
```

and use a feature branch for future PRs (which helps a lot with the GitHub workflow.)

PR #30, opened 2016-06-06
use https everywhere

minrk:
needed to avoid insecure content warnings when loaded on https pages (e.g. jupyter.org)


SylvainCorlay:
:+1:

PR #39, opened 2016-09-26
Jupyterlab Extension

SylvainCorlay:
WIP

leaflet
- [ ] We need to fix the use of private ipywidgets APIs (widget abbreviations)

### DIFF
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -8,9 +8,6 @@ from traitlets import (
     default, validate, TraitError
 )
 
-# TODO: don't use private APIs
-from ipywidgets.widgets.interaction import _widget_from_abbrev
-
 def_loc = [0.0, 0.0]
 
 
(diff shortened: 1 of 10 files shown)


ipyleaflet/leaflet.py
```diff
@@ -8,9 +8,6 @@
     default, validate, TraitError
 )
 
-# TODO: don't use private APIs
-from ipywidgets.widgets.interaction import _widget_from_abbrev
-
```
SylvainCorlay: 💯

SylvainCorlay:
@ellisonbg can you comment on what the InteractMixins were doing? This does not work with more recent versions of ipywidgets and I am not sure what it is supposed to do.

SylvainCorlay:
@jasongrout ping.
This version works with the ipywidgets 6 and jupyterlab 0.7.

SylvainCorlay:
@jasongrout self merging.

PR #66, opened 2017-07-06
Update to ipywidgets 7.0 alpha.

jasongrout:


stonebig:
it still doesn't work on latest jupyterlab, maybe this exact match is the problem ?

PR #73, opened 2017-08-22
updating to newer versions of leaflet

ssunkara1:
@jasongrout


stonebig:
maybe an alpha release with this commit on pypi would help

PR #81, opened 2017-11-06
WIP Add support for basemap

DavideDeMarchi:

PR #87, opened 2017-12-01
Basemap

SylvainCorlay:
This iterates upon #81 , using `Bunch` for auto-completion when choosing the basemap.

Fixes #81 

ipyleaflet

@DavideDeMarchi  in your PR and in this one, the MODIS base map don't seem to be displaying anything. They are the only ones requiring the date as part of the URL.


DavideDeMarchi:
I'm not sure... but the default for MODIS is to display the image for TODAY which is dynamically updated during the day from East to West following the sun, so it could be a black image for part of the globe, depending on current time. Maybe it's better to display the YESTERDAY image if no modisdate parameter is passed.

SylvainCorlay:
OK, I was displaying the maps at a zoom level that was not provided by this service.

PR #89, opened 2017-12-06
Default to using a LayersControl widget

SylvainCorlay:
- Create a `LayersControl` by default.
- Observe the list of available layers to update the layers controls upon addition or removal of layers.

cc @DavideDeMarchi


SylvainCorlay:
@DavideDeMarchi I have release 0.5.2 in the mean time, so that you can already use the layer control, and the basemaps with completion.

SylvainCorlay:
Closing

PR #90, opened 2017-12-14
Add a marker cluster feature

### LINKED ISSUE #63: MarkerCluster?
Hi there,

first of all, thanks for the great library! 

I'm wondering whether there is a way to use a Marker Cluster in `ipyleaflet`, such as is available in `folium` through the `folium.MarkerCluster`?

If not, I wouldn't mind trying to add this functionality and provide a PR, if I can get some pointers as to how to go about implementing it.

Thanks!
Omri

jasongrout:
Fixes #63, using https://github.com/Leaflet/Leaflet.markercluster

### DIFF
--- a/examples/Primitives.ipynb
+++ b/examples/Primitives.ipynb
@@ -12,7 +12,7 @@
     "\n",
     "from ipyleaflet import (\n",
     "    Map,\n",
-    "    Marker,\n",
+    "    Marker, MarkerCluster,\n",
     "    TileLayer, ImageOverlay,\n",
     "    Polyline, Polygon, Rectangle, Circle, CircleMarker,\n",
     "    GeoJSON,\n",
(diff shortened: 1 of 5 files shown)


jasongrout:
@rsignell-usgs, according to https://github.com/Leaflet/Leaflet.markercluster#bulk-adding-and-removing-markers, if we often are removing a lot of markers, we should use a different strategy than what we are doing here with just `removeLayer`. What do you think the typical usecase is? Do you think we would normally just be appending markers to a markercluster, or do you think we would more typically be changing markers in the markercluster?

jasongrout:
I made a change to always use the bulk functions, and forcefully clear everything unless we are in an append-only case.

rsignell-usgs:
In my immediate use case, I would only be modifying characteristics of individual markers in the cluster, not adding or removing.

jasongrout:
FYI, we recorded the initial implementation of this feature, along with lots of comments about how widgets work, at https://www.youtube.com/watch?v=msJig1fr7Lw. (Thanks again, @rsignell-usgs!)

PR #98, opened 2018-02-03
Numpy example

davidbrochart:


SylvainCorlay:
This is a really nice example. Thanks!

Quick note: there are a couple of warnings from rasterio about gdal-style transforms being deprecated. Is it easy to fix?

PR #101, opened 2018-02-04
InvalidateSize upon phosphor resize event

SylvainCorlay:
Calling `InvalidateSize` upon `resize` and `after-show`. It seems that the latter is unnecessary because the resize event is triggered right after.

cc @blink1073

### DIFF
--- a/js/src/jupyter-leaflet.js
+++ b/js/src/jupyter-leaflet.js
@@ -502,12 +502,7 @@ var LeafletMapView = widgets.DOMWidgetView.extend({
         this.leaflet_events();
         this.model_events();
         this.update_bounds();
-        // TODO: hack to get all the tiles to show.
-        var that = this;
-        window.setTimeout(function () {
-            that.obj.invalidateSize();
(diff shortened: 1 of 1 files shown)


js/src/jupyter-leaflet.js
```diff
@@ -583,10 +578,16 @@ var LeafletMapView = widgets.DOMWidgetView.extend({
...
 
-    onResize: function (msg) {
-        this.obj.invalidateSize(true);
-    }
-
+    processPhosphorMessage: function(msg) {
+        LeafletMapView.__super__.processPhosphorMessage.apply(this, arguments);
+            switch (msg.type) {
+                case 'resize':
+                    this.obj.invalidateSize(true);
```
blink1073: missing break statements, this will get called twice

SylvainCorlay: thanks that was it. under my nose!

SylvainCorlay: Fixed.

PR #110, opened 2018-02-09
Fixed map center longitude bug

davidbrochart:


js/src/jupyter-leaflet.js
```diff
@@ -487,7 +487,10 @@ var LeafletMapView = widgets.DOMWidgetView.extend({
     },
 
     render: function () {
-        this.el.style['width'] = this.model.get('width');
+        // Width should not be set by default since it is automatically
+        // stretched in default layout of the map:
+        // return Layout(height='400px', align_self='stretch')
+        //this.el.style['width'] = this.model.get('width');
```
SylvainCorlay: We should probably call `super.render();`and not get the height or the width at all?

davidbrochart: Not sure how all that works to be honest. I just saw that the JS side can't assume the width will be 600px since in the Python side we say it will fit the cell.

SylvainCorlay: We need to completely remove `width` and `height` information from the LeafletMap model since it is all handled by the base widgets logic.

SylvainCorlay: In the past, it was not the case, this is leftover from when ipyleaflet worked on the top of ipywidgets 4.x.

davidbrochart: The strange thing is that if I do the same with the height (remove line 491) it doesn't work.

SylvainCorlay: OK, I will check it out and comment further here.

SylvainCorlay: So it works with removing all references to `width` and `height` if you do

```js
        LeafletMapView.__super__.render.apply(this)
```

in the beginning of the render method.

davidbrochart: Not on my side. Typically what I do is:
```python
center = [-10, -50]
zoom = 2
m = Map(center=center, zoom=zoom)
m
```
That should display a map centered on the Amazon. In that case it doesn't. If it does, try executing the cell several times. Then `m.center` gives `[-10, -50]`, but if you slightly move the map with the mouse it gives something like `[-62.26792262941758, -50.27343750000001]`.

SylvainCorlay: I think that it is a race condition between the first invalidateSize and the resizing of the map..

davidbrochart: Then I don't understand why with my fix it never happens.

SylvainCorlay:
This is now fixed in master.

PR #118, opened 2018-02-14
Prevent subpixel leaflet errors to impact model values

SylvainCorlay:
- Use a `dirty` flag on the view to prevent sub-pixel center changes computed by leaflet.js to be applied to the model.
 - Use `flyTo` instead of `setZoom` to apply model's center value upon change of zoom. Indeed, subpixel errors at low zoom level may correspond to large error at the new higher zoom level.


SylvainCorlay:
cc @davideDeMarchi

PR #131, opened 2018-03-19
Update CustomTMS examples use of default tiles.

tylere:
Removes the use of Map.default_tiles which was removed in Release 0.7.0.


SylvainCorlay:
LGTM!

SylvainCorlay:
Thanks

PR #132, opened 2018-03-20
Fix race condaition for double-click zoom

### LINKED ISSUE #130: After double click zoom map may not contain area that was clicked
When zooming by using a double click, the center of the map stays constant, rather than panning to include the area where the user double clicked. This seems non-intuitive, and is in contrast to the default Leaflet behavior (as demonstrated on the [Leaflet website](http://leafletjs.com/)), which pans the map to include the area where the user has double clicked.

SylvainCorlay:
Fixes #130

### DIFF
--- a/js/src/jupyter-leaflet.js
+++ b/js/src/jupyter-leaflet.js
@@ -665,17 +665,19 @@ var LeafletMapView = widgets.DOMWidgetView.extend({
         var that = this;
         this.obj.on('moveend', function (e) {
             if (!that.dirty) {
+                that.dirty = true;
                 var c = e.target.getCenter();
                 that.model.set('center', [c.lat, c.lng]);
-                that.touch();
+                that.dirty = false;
             }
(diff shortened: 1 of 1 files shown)


Juanlu001:
I tried a development installation but I think this is wrong:

https://github.com/SylvainCorlay/ipyleaflet/blob/fff7e0c6f2a86e91d449f67252728b9129cf387b/js/package-lock.json#L2725

That commit does not belong to any branch:

https://github.com/QuantStack/leaflet-splitmap/commit/ac1b55f001d0366768a0fe1cf6b94a0197c1d082

Juanlu001:
It's actually this commit https://github.com/QuantStack/leaflet-splitmap/commit/f9761db56f710e21eb3bc809d765ce3b1bbdfbe3

Juanlu001:
After fixing the commit and performing a development installation, the double click zoom works correctly for me :+1:

PR #141, opened 2018-04-07
Fix Circle and CircleMarker classes inheritance

### LINKED ISSUE #140: Circle and CircleMarker classes
Circle class inherits from Path and CircleMarker class inherits from Circle, which is not how leaflet is designed:

https://github.com/Leaflet/Leaflet/blob/0b868472a22e5dc190ff2df9135f6d73b2366120/src/layer/vector/Circle.js#L25

and 

https://github.com/Leaflet/Leaflet/blob/0b868472a22e5dc190ff2df9135f6d73b2366120/src/layer/vector/CircleMarker.js#L15

martinRenou:
Fixes #140 
Fixes inheritance of Circle and CircleMarker classes according to leaflet implementation

### DIFF
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -432,19 +432,19 @@ class Rectangle(Polygon):
     bounds = List(help="list of SW and NE location tuples").tag(sync=True)
 
 
-class Circle(Path):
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
LGTM!

PR #155, opened 2018-04-12
Fix moveend event for map

martinRenou:


martinRenou:
This event should not be sent to the model

PR #166, opened 2018-04-24
Heatmap

davidbrochart:
This uses the Leaflet heatmap plugin:
- https://github.com/Leaflet/Leaflet.heat
- https://github.com/mourner/simpleheat

### DIFF
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -212,15 +212,14 @@ class Heatmap(RasterLayer):
     _model_name = Unicode('LeafletHeatmapModel').tag(sync=True)
 
     latlngs = List().tag(sync=True)
-    minOpacity = Float(0.05).tag(sync=True)
-    maxZoom = Int(18).tag(sync=True)
-    maxIntensity = Float(1.0).tag(sync=True)
-    radius = Float(25.0).tag(sync=True)
-    blur = Float(15.0).tag(sync=True)
-    gradient = Dict({0.4: 'blue', 0.6: 'cyan', 0.7: 'lime', 0.8: 'yellow', 1.0: 'red'}).tag(sync=True)
 
     # Options
-    attribution = Unicode().tag(sync=True, o=True)
+    minOpacity = Float(0.05).tag(sync=True, o=True)
+    maxZoom = Int(18).tag(sync=True, o=True)
+    max = Float(1.0).tag(sync=True, o=True)
+    radius = Float(25.0).tag(sync=True, o=True)
+    blur = Float(15.0).tag(sync=True, o=True)
+    gradient = Dict({0.4: 'blue', 0.6: 'cyan', 0.7: 'lime', 0.8: 'yellow', 1.0: 'red'}).tag(sync=True, o=True)
 
 
 class VectorLayer(Layer):
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -207,6 +207,22 @@ class VideoOverlay(RasterLayer):
     attribution = Unicode().tag(sync=True, o=True)
 
 
+class Heatmap(RasterLayer):
+    _view_name = Unicode('LeafletHeatmapView').tag(sync=True)
+    _model_name = Unicode('LeafletHeatmapModel').tag(sync=True)
+
+    latlngs = List().tag(sync=True)
+    minOpacity = Float(0.05).tag(sync=True)
+    maxZoom = Int(18).tag(sync=True)
+    maxIntensity = Float(1.0).tag(sync=True)
+    radius = Float(25.0).tag(sync=True)
+    blur = Float(15.0).tag(sync=True)
+    gradient = Dict({0.4: 'blue', 0.6: 'cyan', 0.7: 'lime', 0.8: 'yellow', 1.0: 'red'}).tag(sync=True)
+
+    # Options
+    attribution = Unicode().tag(sync=True, o=True)
+
+
 class VectorLayer(Layer):
     _view_name = Unicode('LeafletVectorLayerView').tag(sync=True)
     _model_name = Unicode('LeafletVectorLayerModel').tag(sync=True)
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -214,8 +214,8 @@ class Heatmap(RasterLayer):
     latlngs = List().tag(sync=True)
 
     # Options
-    minOpacity = Float(0.05).tag(sync=True, o=True)
-    maxZoom = Int(18).tag(sync=True, o=True)
+    min_opacity = Float(0.05).tag(sync=True, o=True)
+    max_zoom = Int(18).tag(sync=True, o=True)
     max = Float(1.0).tag(sync=True, o=True)
     radius = Float(25.0).tag(sync=True, o=True)
     blur = Float(15.0).tag(sync=True, o=True)
--- a/js/src/jupyter-leaflet.js
+++ b/js/src/jupyter-leaflet.js
@@ -364,8 +364,6 @@ var LeafletHeatmapView = LeafletLayerView.extend({
 
     model_events: function () {
         LeafletHeatmapView.__super__.model_events.apply(this, arguments);
-        this.listenTo(this.model, 'change:latlngs', function () {
-        }, this);
 
         this.listenTo(this.model, 'change:options', function () {
             latlngs = this.model.get('latlngs');
@@ -377,10 +375,7 @@ var LeafletHeatmapView = LeafletLayerView.extend({
                 blur: this.model.get('blur'),
                 gradient: this.model.get('gradient')
             }
-            options = this.get_options();
-            this.map_view.obj.removeLayer(this.obj);
-            this.obj = L.heatLayer(latlngs, heat_options, options);
-            this.map_view.obj.addLayer(this.obj);
+            this.obj.setOptions(this.get_options());
         }, this);
     },
 });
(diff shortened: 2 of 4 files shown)


SylvainCorlay:
Thanks! I am posting a couple of questions online!

js/src/jupyter-leaflet.js
```diff
@@ -4,6 +4,11 @@ var L = require('leaflet');
 require('leaflet-splitmap');
 require('leaflet-draw');
 require('leaflet.markercluster');
+Heat = require('./leaflet-heat.js');
+L.HeatLayer = (L.Layer ? L.Layer : L.Class).extend(Heat);
```
SylvainCorlay: Is this a compatibility thing between several versions of leaflet?

davidbrochart: I don't know, this comes from the original code:
https://github.com/Leaflet/Leaflet.heat/blob/gh-pages/src/HeatLayer.js#L3

SylvainCorlay:
Thanks! I am off for a few days, but will get back to this as soon as I am back.

js/src/jupyter-leaflet.js
```diff
@@ -344,6 +345,45 @@ var LeafletVideoOverlayView = LeafletRasterLayerView.extend({
...
+                gradient: this.model.get('gradient')
+            },
+            this.get_options()
+        );
+    },
+
+    model_events: function () {
+        LeafletHeatmapView.__super__.model_events.apply(this, arguments);
+        this.listenTo(this.model, 'change:latlngs', function () {
+        }, this);
```
martinRenou: You don't do anything in the function, don't you need a
```javascript
this.obj.setLatLngs(this.model.get('latlngs'));
```
?

js/src/jupyter-leaflet.js
```diff
@@ -1069,6 +1109,16 @@ var LeafletVideoOverlayModel = LeafletRasterLayerModel.extend({
     })
 });
 
+var LeafletHeatmapModel = LeafletRasterLayerModel.extend({
+    defaults: _.extend({}, LeafletRasterLayerModel.prototype.defaults, {
+        _view_name : 'LeafletHeatmapView',
+        _model_name : 'LeafletHeatmapModel',
+
+        latlngs : [],
+        attribution : ''
```
martinRenou: You could also add the default values for the other attributes
```javascript
minOpacity: 0.05,
maxZoom: 18,
maxIntensity: 1.0,
radius: 25.0,
blur: 15.0,
gradient: {0.4: 'blue', 0.6: 'cyan', 0.7: 'lime', 0.8: 'yellow', 1.0: 'red'}
```

js/src/leaflet-heat.js
```diff
@@ -0,0 +1,379 @@
...
+
+L.HeatLayer = (L.Layer ? L.Layer : L.Class).extend({
+
+    // options: {
+    //     minOpacity: 0.05,
+    //     maxZoom: 18,
+    //     radius: 25,
+    //     blur: 15,
+    //     max: 1.0
+    // },
```
martinRenou: Maybe the options should not be commented out ?

js/src/jupyter-leaflet.js
```diff
@@ -344,6 +345,45 @@ var LeafletVideoOverlayView = LeafletRasterLayerView.extend({
...
+                maxZoom: this.model.get('maxZoom'),
+                max: this.model.get('maxIntensity'),
+                radius: this.model.get('radius'),
+                blur: this.model.get('blur'),
+                gradient: this.model.get('gradient')
+            }
+            options = this.get_options();
+            this.map_view.obj.removeLayer(this.obj);
+            this.obj = L.heatLayer(latlngs, heat_options, options);
+            this.map_view.obj.addLayer(this.obj);
```
martinRenou: Why do you remove the layer and then recreate it? Something like
```javascript
this.obj.setOptions(this.get_options());
```
should do the job

davidbrochart:
Hi @martinRenou, I took you suggestions in account in the latest commit. Thanks!

martinRenou:
Cool :) thanks

ipyleaflet/leaflet.py
```diff
@@ -207,6 +207,22 @@ class VideoOverlay(RasterLayer):
...
+    _view_name = Unicode('LeafletHeatmapView').tag(sync=True)
+    _model_name = Unicode('LeafletHeatmapModel').tag(sync=True)
+
+    latlngs = List().tag(sync=True)
+    minOpacity = Float(0.05).tag(sync=True)
+    maxZoom = Int(18).tag(sync=True)
+    maxIntensity = Float(1.0).tag(sync=True)
+    radius = Float(25.0).tag(sync=True)
+    blur = Float(15.0).tag(sync=True)
+    gradient = Dict({0.4: 'blue', 0.6: 'cyan', 0.7: 'lime', 0.8: 'yellow', 1.0: 'red'}).tag(sync=True)
```
martinRenou: Sorry... I didn't see it yesterday, `minOpacity`, `maxZoom`, `max`, `radius`, `blur`, `gradient` are options. So you need to specify it with `o=True` like:
```python
blur = Float(15.0).tag(sync=True, o=True)
```
Otherwise you wont be able to update those options from the python side:
```python
heat.blur = 5  # Does not work currently
```

ipyleaflet/leaflet.py
```diff
@@ -207,6 +207,22 @@ class VideoOverlay(RasterLayer):
...
+
+    # Options
+    attribution = Unicode().tag(sync=True, o=True)
```
martinRenou: I'm not sure, but I think this can be removed. You don't seem to use it on the javascript side.

js/src/jupyter-leaflet.js
```diff
@@ -344,6 +345,40 @@ var LeafletVideoOverlayView = LeafletRasterLayerView.extend({
...
+            },
+            this.get_options()
+        );
+    },
+
+    model_events: function () {
+        LeafletHeatmapView.__super__.model_events.apply(this, arguments);
+
+        this.listenTo(this.model, 'change:options', function () {
+            latlngs = this.model.get('latlngs');
```
martinRenou: Unused here, you can remove it. You need a special event handler for the `latlngs`, because it's not an option.

js/src/jupyter-leaflet.js
```diff
@@ -344,6 +345,40 @@ var LeafletVideoOverlayView = LeafletRasterLayerView.extend({
...
+                blur: this.model.get('blur'),
+                gradient: this.model.get('gradient')
+            },
+            this.get_options()
+        );
+    },
+
+    model_events: function () {
+        LeafletHeatmapView.__super__.model_events.apply(this, arguments);
+
```
martinRenou: Your event handler for `latlngs` can be something like:
```javascript
this.listenTo(this.model, 'change:latlngs', function () {
    this.obj.setLatLngs(this.model.get('latlngs'));
}, this);
```

js/src/jupyter-leaflet.js
```diff
@@ -1069,6 +1104,22 @@ var LeafletVideoOverlayModel = LeafletRasterLayerModel.extend({
     })
 });
 
+var LeafletHeatmapModel = LeafletRasterLayerModel.extend({
+    defaults: _.extend({}, LeafletRasterLayerModel.prototype.defaults, {
+        _view_name : 'LeafletHeatmapView',
+        _model_name : 'LeafletHeatmapModel',
+
+        latlngs : [],
+        attribution : '',
```
martinRenou: Can also be removed

davidbrochart:
Should be okay now.

martinRenou:
Awesome! :)

ipyleaflet/leaflet.py
```diff
@@ -207,6 +207,21 @@ class VideoOverlay(RasterLayer):
...
 
+class Heatmap(RasterLayer):
+    _view_name = Unicode('LeafletHeatmapView').tag(sync=True)
+    _model_name = Unicode('LeafletHeatmapModel').tag(sync=True)
+
+    latlngs = List().tag(sync=True)
+
+    # Options
+    minOpacity = Float(0.05).tag(sync=True, o=True)
+    maxZoom = Int(18).tag(sync=True, o=True)
```
SylvainCorlay: `ipyleaflet` automatically converts `snake_case` to `camelCase` between the front-end and the back-end, so normally you should be able to call these traits `min_opacity` and `max_zoom` without any other change.

davidbrochart: Yes I discovered that recently. Any reason we have `snake_case` on the Python side and `camelCase` on the JS side, and not the same names on both sides?

SylvainCorlay: Mostly because of PEP8.

SylvainCorlay:
OK, I think that we are good to merge.

I wanted to put in a few remarks for the record. Maybe we should address some of them before a release

 - is there a node package for the heatmap extension? Should we depend on it instead of vendoring the file?
 - `latlngs` may not be the best name for the point locations.
 - should we migrate to a binary serialization for the `latlngs` array or at depend on `traittypes` to make it an `Array` trait?

martinRenou:
There is a [node package](https://www.npmjs.com/package/leaflet-heatmap), but it's old. Actually this heatmap extension has not been updated since a long time ago.

`latlngs` could be renamed to `locations`, just to stay consistent with `Marker` and `Popup`.

SylvainCorlay:
I would be +1 on `locations`.

davidbrochart:
- There is no node package for this heatmap plugin, the one @martinRenou mentioned is for [heatmap.js](https://github.com/pa7/heatmap.js). Maybe I made the wrong choice at the beginning...
- `latlngs` comes from [Leaflet](https://leafletjs.com/reference-1.3.0.html#latlng), but it's okay to rename it.
- Not sure what the differences are between binary serialization and `Array` trait, I guess binary is more efficient for lots of points?

SylvainCorlay:
Ok. Merging as is. I think that we should tackle the other questiosn in another PR.

SylvainCorlay:
FYI, I did the renaming to locations and added a section in the documentation.

PR #171, opened 2018-05-20
Numpy example update

davidbrochart:
Update with latest Rasterio (RESAMPLING -> Resampling).
Add image interpolation option to support nearest neighbor mode.


SylvainCorlay:
Thanks! Merging.

PR #178, opened 2018-06-10
Add leaflet-velocity layer

davidbrochart:
The velocity data is just read from a JSON file. It might be worth looking at https://github.com/benbovy/xvelmap for xarray integration.

### DIFF
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -207,6 +207,45 @@ class VideoOverlay(RasterLayer):
     attribution = Unicode().tag(sync=True, o=True)
 
 
+class Velocity(Layer):
+    _view_name = Unicode('LeafletVelocityView').tag(sync=True)
+    _model_name = Unicode('LeafletVelocityModel').tag(sync=True)
+
+    data = Unicode().tag(sync=True)
+
+    # Options
+    display_values = Bool(True).tag(sync=True, o=True)
+    display_options = Dict({
+        'velocityType': 'Global Wind',
+        'position': 'bottomleft',
+        'emptyString': 'No velocity data',
+        'angleConvention': 'bearingCW',
+        'displayPosition': 'bottomleft',
+        'displayEmptyString': 'No velocity data',
+        'speedUnit': 'kt'
+        }).tag(sync=True, o=True)
+    min_velocity = Float(0).tag(sync=True,o=True)
+    max_velocity = Float(10).tag(sync=True, o=True)
+    velocity_scale = Float(0.005).tag(sync=True, o=True)
+    color_scale = List([
+        "rgb(36,104, 180)",
+        "rgb(60,157, 194)",
+        "rgb(128,205,193)",
+        "rgb(151,218,168)",
+        "rgb(198,231,181)",
+        "rgb(238,247,217)",
+        "rgb(255,238,159)",
+        "rgb(252,217,125)",
+        "rgb(255,182,100)",
+        "rgb(252,150,75)",
+        "rgb(250,112,52)",
+        "rgb(245,64,32)",
+        "rgb(237,45,28)",
+        "rgb(220,24,32)",
+        "rgb(180,0,35)"
+        ]).tag(sync=True, o=True)
+
+
 class VectorLayer(Layer):
     _view_name = Unicode('LeafletVectorLayerView').tag(sync=True)
     _model_name = Unicode('LeafletVectorLayerModel').tag(sync=True)
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -8,9 +8,17 @@ from traitlets import (
     default, validate, TraitError
 )
 
+from .xarray_ds import ds2json
+
 from .basemaps import basemaps
 
 from ._version import EXTENSION_VERSION
+
+try:
+    from xarray import Dataset
+except:
+    class Dataset:
+        pass
 
 def_loc = [0.0, 0.0]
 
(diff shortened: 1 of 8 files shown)


js/src/jupyter-leaflet.js
```diff
@@ -344,6 +345,41 @@ var LeafletVideoOverlayView = LeafletRasterLayerView.extend({
...
+
+    model_events: function () {
+        LeafletVelocityView.__super__.model_events.apply(this, arguments);
+        this.listenTo(this.model, 'change:data', function () {
+            options = this.get_options();
+            data = this.model.get('data');
+            options.data = JSON.parse(data);
+            this.map_view.obj.removeLayer(this.obj);
+            this.obj = L.velocityLayer(options);
+            this.map_view.obj.addLayer(this.obj);
```
martinRenou: There is a `setData` method that you can use instead of removing and adding back the layer :)

js/src/jupyter-leaflet.js
```diff
@@ -344,6 +345,41 @@ var LeafletVideoOverlayView = LeafletRasterLayerView.extend({
...
+        var o = this.model.get('options');
+        for (var i=0; i<o.length; i++) {
+            key = o[i];
+            this.listenTo(this.model, 'change:' + key, function () {
+                options = this.get_options();
+                data = this.model.get('data');
+                options.data = JSON.parse(data);
+                this.map_view.obj.removeLayer(this.obj);
+                this.obj = L.velocityLayer(options);
+                this.map_view.obj.addLayer(this.obj);
```
martinRenou: If you use `L.setOptions(this.obj, options)` instead of removing and adding back the layer, I guess it should work, but I'm really not sure for this one

js/src/jupyter-leaflet.js
davidbrochart: Right.

js/src/jupyter-leaflet.js
davidbrochart: Yes it works, thanks!

davidbrochart:
The last commit steals code from [xvelmap](https://github.com/benbovy/xvelmap/blob/master/xvelmap/xr_accessor.py). It allows to get data from an xarray dataset (it is converted to JSON and fed to leaflet-velocity).

ipyleaflet/leaflet.py
```diff
@@ -211,8 +219,19 @@ class Velocity(Layer):
...
+    lat_dim = Unicode('latitude')
+    lon_dim = Unicode('longitude')
+    units = Unicode(None, allow_none=True)
+
+    dataset = Instance(Dataset)
     data = Unicode().tag(sync=True)
 
+    def __init__(self, **kwargs):
+        super(Velocity, self).__init__(**kwargs)
+        self.set_trait('data', ds2json(self.dataset, self.u_var, self.v_var, self.lat_dim, self.lon_dim))
```
SylvainCorlay: Another way to do this is to not have

```
data = Instance(Dataset).tag(sync=True, to_json=ds_x_to_json, from_json=ds_x_from_json)
```

where `ds_x_to_json` essentially wraps `ds2json`.

SylvainCorlay: You would mostly need to remove the `json.dumps` part of the function, which could become private.

davidbrochart: Great, I didn't know. I'm updating it.

davidbrochart: Also, there is a warning because traitlets is trying to compare datasets to detect changes.

SylvainCorlay: Yeah, the trick would be to add `DataSet` to https://github.com/jupyter-widgets/traittypes

SylvainCorlay: (Where we also overrode the set to not use `==`).

SylvainCorlay: Traittypes would become a dependency to ipyleaflet. It is already a dependency of pythreejs, ipyvolume and bqplot...

davidbrochart: And I was not able to observe `dataset` to update `data`, probably because of this comparison issue.

ipyleaflet/xarray_ds.py
```diff
@@ -0,0 +1,108 @@
...
+                "lo2": lon_right,
+                "nx": nx,
+                "ny": ny,
+                "refTime": "2017-02-01 23:00:00",
+                "lo1": lon_left
+                },
+            "data": ds[var_name].values.flatten().tolist()
+        })
+
+    return json.dumps(velocity_data)
```
SylvainCorlay: You probably need to remove the `json.dumps`

davidbrochart: No, it looks like it's needed.

SylvainCorlay: Not if you  remove the JSON.parse in `options.data = JSON.parse(data)`.

davidbrochart: Right!

SylvainCorlay:
Could you check that traittypes was doing the job with your latest additions?

If it is the case, I can push out a release so that you can move forward here.

davidbrochart:
Sure but probably not today, sorry for that.

SylvainCorlay:
> Sure but probably not today, sorry for that.

Well, I have not said it here, but that PR is *awesome*. When it gets in, it will be really really nice for ipyleaflet.

davidbrochart:
Thanks!

davidbrochart:
Good to go with future Dataset traittypes! The only thing is that there is a warning from xarray when passing a Dataset object to the Dataset constructor (see https://github.com/pydata/xarray/issues/2056).

SylvainCorlay:
OK, traittypes 0.2.1 is out!

The conda-forge package is building.

SylvainCorlay:
OK, I could test things again! Merging.

Thanks again, this is pretty awesome.

Putting up a screen cast:

velocity

PR #181, opened 2018-06-12
Update image/video layer upon URL change.

ian-r-rose:
I have a use-case where I wanted to dynamically update the image url of a layer, which seemed to be missing. Any reason not to add it here?


SylvainCorlay:
Thanks!
Indeed no reason to not have it. This should get in before the next release with the heatmap and velocity layer!

SylvainCorlay:
OK, I could test it and it appears to work.

Although I am not seeing the warning that you are describing.

ian-r-rose:
Yeah, I never saw that warning either, just a copy-paste oversight. Thanks!

PR #183, opened 2018-06-17
Add MeasureControl

martinRenou:


martinRenou:
This doesn't work yet, the CSS seems to be not correctly loaded. `leaflet-measure` package uses `scss` files

martinRenou:
@SylvainCorlay should be fine now, I had to use some workarounds, but nothing too ugly I think

js/src/jupyter-leaflet.js
```diff
@@ -705,6 +706,51 @@ var LeafletLayersControlView = LeafletControlView.extend({
...
+    model_events: function () {
+        var key;
+        var o = this.model.get('options');
+        for (var i=0; i<o.length; i++) {
+            key = o[i];
+            this.listenTo(this.model, 'change:' + key, function () {
+                // Workaround for https://github.com/ljagis/leaflet-measure/issues/112
+                // and https://github.com/ljagis/leaflet-measure/issues/113
+                // Once fixed, the next line should be replaced by: L.setOptions(this.obj, this.get_options());
+                this.obj.initialize(this.get_options());
```
SylvainCorlay: 👍

js/src/jupyter-leaflet.js
```diff
@@ -705,6 +706,51 @@ var LeafletLayersControlView = LeafletControlView.extend({
...
+
+    get_options: function () {
+        var o = this.model.get('options');
+        var options = {};
+        var key;
+        for (var i=0; i<o.length; i++) {
+            key = o[i];
+            // Convert from foo_bar to fooBar that Leaflet.js uses
+            options[camel_case(key)] = this.model.get(key);
+        }
```
SylvainCorlay: maybe call super.get_options instead of duplicating the implementation.

martinRenou: There is no such method in ControlView, but I will add it. Maybe it will factorize some code 😊

js/src/jupyter-leaflet.js
```diff
@@ -27,8 +28,24 @@ function camel_case(input) {
...
+        return options;
+    }
+}
 
-var LeafletLayerView = widgets.WidgetView.extend({
+var LeafletWidgetView = widgets.WidgetView.extend(leaflet_views_common_methods);
+var LeafletDOMWidgetView = widgets.DOMWidgetView.extend(leaflet_views_common_methods);
```
SylvainCorlay: Well done. Good idea to use mixins.

martinRenou: Thanks :)

Issue #173, opened 2018-05-30
Failed to display Jupyter Widget of type Map


eRajsh:
Had created an issue under JupyterLab repo (https://github.com/jupyterlab/jupyterlab/issues/4654). Maybe this is the right place.

Recently installed Anaconda 3.5.1 on mac, and trying to write a simple map thingy in the notebook.

Ran the commands (pip install ipyleaflet) successfully

Extensions seem to be properly installed:
![](https://user-images.githubusercontent.com/1895122/40637836-33059402-6325-11e8-9190-4a9c706a0c74.png)

Here is the code in the notebook:
***************************************************************
!jupyter nbextension enable --py widgetsnbextension --sys-prefix
!jupyter nbextension enable --py --sys-prefix ipyleaflet

import pandas as pd
from ipyleaflet import Map

Map(center=[ 37.7749, -112.4194], zoom = 10)
***************************************************************

When I run the notebook, I get the following issue:
Enabling notebook extension jupyter-js-widgets/extension...
- Validating: OK
Enabling notebook extension jupyter-leaflet/extension...
- Validating: OK
Failed to display Jupyter Widget of type Map.

clemens-tolboom:
I get the same report but succesfull map

```bash
> jupyter nbextension list
Known nbextensions:
  config dir: /Users/.../bin/../etc/jupyter/nbconfig
    notebook section
      jupyter-leaflet/extension  enabled 
      - Validating: OK
      jupyter-js-widgets/extension  enabled 
      - Validating: OK
```

Not sure this is useful but pasting the code into a notebook

```python
import pandas as pd
from ipyleaflet import Map

Map(center=[ 37.7749, -112.4194], zoom = 10)
```
produces a map.

I'm on MacOs using `pipenv`

eRajsh:
Thx Clemens-tolboom for trying it out. I cannot resolve the issue, tried reinstalling from scratch, still the same. I am also on Mac.

SylvainCorlay:
@eRajsh do you see errors in the js console (option/command/j to open the developer tools on mac).

SylvainCorlay:
Also, can you please provide the full output of `jupyter nbextension list`?

costrouc:
I have the exact same issue. I created a new virtualenv to test. Here are the exact steps that I took on Ubuntu 18.04. 

```bash
pipenv install ipyleaflet jupyterlab
pipenv run jupyter labextension install jupyter-leaflet
[...]

Issue #168, opened 2018-05-03
jupyter lab no graphical output, just prints repr


mfortini:
I installed jupyter, jupyter labs and ipyleaflet using pip in a clean venv

A cell with this code:
```
from ipyleaflet import *
m = Map(center=(52, 10), zoom=8, basemap=basemaps.Hydda.Full)
m
```
is not showing any graphical output, just
```
Map(basemap={'url': 'http://{s}.tile.openstreetmap.se/hydda/full/{z}/{x}/{y}.png', 'max_zoom': 18, 'attributio…
```

The package seems installed ok:
```
jupyter labextension list
JupyterLab v0.32.1
Known labextensions:
   app dir: [...]/jupyter/share/jupyter/lab
jupyter-leaflet
        jupyter-leaflet v0.8.1  enabled  OK
```

SylvainCorlay:
Thanks for reporting.

Apparently you need to enable the base widgetw nbextension.

```
jupyter labextension install @jupyter-widgets/jupyterlab-manager
```

mfortini:
Great, my bad, it works now. Thank you.

Asr-ar:
still not working for me ('

PR #190, opened 2018-06-27
Adding documentation about getting jupyterlab ipyleaflet working

costrouc:
I ran into an issue that ipyleaflet would only display in `jupyter
notebooks` and not `jupyter lab`. This details the additional step
needed to get it working on Jupyterlab.

This addresses these two issues #168  and #173


SylvainCorlay:
Thanks for updating the readme!

PR #191, opened 2018-06-27
Switch to jupyterlab for binder examples

costrouc:
I previously submitted a pull request for improving the documentation on using ipyleaflet with jupyterlab. In fact I needed it to work for a binder example project of my own. So I wanted to offer a minimal pull request that will set the default binder launch button to use jupyterlab instead of jupyter notebooks. 

The only addition required is a [postBuild](https://mybinder.readthedocs.io/en/latest/using.html#postbuild) file for adding the two labextensions: jupyter-manager and ipyleaflet.  Description of what a postBuild file is [here](https://mybinder.readthedocs.io/en/latest/using.html#postbuild)

You may use the following link to test of the binder jupyterlab notebook examples before the pull request. I have already tested that all the example notebooks run and display properly in binder before the request.

https://mybinder.org/v2/gh/costrouc/ipyleaflet/costrouc/binder-jupyterlab?urlpath=lab%2Ftree%2Fexamples

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # ipyleaflet
 
 [![Documentation](http://readthedocs.org/projects/ipyleaflet/badge/?version=latest)](https://ipyleaflet.readthedocs.io/en/latest/?badge=latest)
-[![Binder](https://img.shields.io/badge/launch-binder-brightgreen.svg)](http://mybinder.org/v2/gh/jupyter-widgets/ipyleaflet/0.8.4?urlpath=lab/tree/examples)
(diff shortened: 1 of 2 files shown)


costrouc:
I must also add that I don't think it is worth accepting the pull request for the until the binder link is changed to be appropriate since it won't work as is. Obviously the git tag 0.8.3 wont have the `postBuild` file so it will fail.

SylvainCorlay:
Thanks!

The way I would go with this is
 - only adding the postBuild as part of the PR
 - update the URL in the commit following a release

PR #196, opened 2018-07-03
Rename velocity speed attributes

martinRenou:


martinRenou:
I also need to update the doc

PR #206, opened 2018-07-31
Change example longitudes to more reasonable values inside (-180, 180)

deeplook:
Simple change in longitude values to make them in the more expected range between -180 and 180 degrees.


martinRenou:
Thanks! :)

PR #220, opened 2018-09-10
Rotate markers, Marker Icons

gregfreeman:
How does this look?
Its my first time contributing to jupyter-widgets. 

Changes
------------------
* Added capability to rotate markers with leaflet plugin  https://github.com/bbecquet/Leaflet.RotatedMarker
* Added a widget model and view for marker icons.
* Updated documentation for new features.

### DIFF
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -249,8 +249,8 @@ class Velocity(Layer):
         'displayPosition': 'bottomleft',
         'displayEmptyString': 'No velocity data',
         'speedUnit': 'kt'
-    }).tag(sync=True)
-    min_velocity = Float(0).tag(sync=True, o=True)
+        }).tag(sync=True)
+    min_velocity = Float(0).tag(sync=True,o=True)
     max_velocity = Float(10).tag(sync=True, o=True)
     velocity_scale = Float(0.005).tag(sync=True, o=True)
     color_scale = List([
@@ -269,7 +269,7 @@ class Velocity(Layer):
         "rgb(237,45,28)",
         "rgb(220,24,32)",
         "rgb(180,0,35)"
-    ]).tag(sync=True, o=True)
+        ]).tag(sync=True, o=True)
 
 
 class Heatmap(RasterLayer):
@@ -516,8 +516,8 @@ class MeasureControl(Control):
     completed_color = Color('#C8F2BE').tag(sync=True, o=True)
 
     popup_options = Dict({
-        'className': 'leaflet-measure-resultpopup',
-        'autoPanPadding': [10, 10]
+      'className': 'leaflet-measure-resultpopup',
+      'autoPanPadding': [10, 10]
     }).tag(sync=True, o=True)
 
     capture_z_index = Int(10000).tag(sync=True, o=True)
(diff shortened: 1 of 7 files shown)


js/src/jupyter-leaflet.js
```diff
@@ -160,10 +202,27 @@ var LeafletMarkerView = LeafletUILayerView.extend({
...
 
 
+var LeafletIconView = LeafletUILayerView.extend({
+
+    create_obj: function () {
+        var that = this;
+
+        this.obj = L.icon(this.get_options());
+    },
+
```
martinRenou: Does it work to change icon options dynamically? I mean, if you try something like:
```javascript
this.model.on('change:icon_url', () => {
});
```
Will the icon url be updated for each marker that uses this icon?

martinRenou: I guess the answer is no... I took a look at the implementation of icons in LeafletJs, and I didn't see a way to update the icon options once it has been created. I think the only way to have a dynamic change is to remove the icon and create it again. We have the same problem with popups

gregfreeman: No the icon options can't be changed dynamically, but the marker can be assigned a different icon object  dynamically.   I don't see a setOptions method in leaflet.

ipyleaflet/leaflet.py
```diff
@@ -99,6 +112,9 @@ class Marker(UILayer):
     alt = Unicode('').tag(sync=True, o=True)
     rise_on_hover = Bool(False).tag(sync=True, o=True)
     rise_offset = Int(250).tag(sync=True, o=True)
+    rotation_angle = Float(0).tag(sync=True, o=True)
+    rotation_origin = Unicode('').tag(sync=True, o=True)
+    icon = Instance(Icon, allow_none=True, default_value=None).tag(sync=True, o=True, **widget_serialization)
```
martinRenou: I think you can remove the `o=True`, because on the JavaScript part you remove `icon` from the options object

js/src/jupyter-leaflet.js
```diff
@@ -113,11 +114,20 @@ var LeafletUILayerView = LeafletLayerView.extend({
...
+
     create_obj: function () {
         var that = this;
+        var options = this.get_options();
+        delete options['icon'];
         this.obj = L.marker(
             this.model.get('location'),
-            this.get_options()
+            options
```
martinRenou: If you remove `o=True` from the icon attr in the Python model you can undo those changes.

js/src/jupyter-leaflet.js
```diff
@@ -128,6 +138,32 @@ var LeafletMarkerView = LeafletUILayerView.extend({
...
+        if (this.icon) {
+            this.icon.remove();
+        }
+        if (value) {
+            var that = this;
+            this.icon_promise = this.icon_promise.then(function() {
+                return that.create_child_view(value).then(function(view) {
+                    that.obj.setIcon(view.obj);
+                    that.icon = view;
+                    that.trigger('icon:created');
```
martinRenou: I'm not sure I understand this line, why do you trigger this event? Is it used somewhere else?

SylvainCorlay:
Thanks very much for the contribution! Very comprehensive and including examples and documentation.

Waiting on the review items to be fixed for merging.

gregfreeman:
@martinRenou, Thanks for the feedback.  I've made the fixes.

martinRenou:
Thanks :)

PR #221, opened 2018-10-02
Xarray no longer supports python 3.4

drnextgis:
Since 0.10.8 xarray no longer supports python 3.4, see this [commit](https://github.com/drnextgis/xarray/commit/1688a59803786a9d88eeb43aa4c935f7052d6a80#diff-2eeaed663bd0d25b7e608891384b7298).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -136,7 +136,7 @@ setup_args = {
     'install_requires': [
         'ipywidgets>=7.0.0,<8',
         'traittypes>=0.2.1,<3',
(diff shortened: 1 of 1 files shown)


drnextgis:
Moreover xarray doesn't support python 3.3 since 0.9.0 but we have python 3.3 in our setup.py https://github.com/jupyter-widgets/ipyleaflet/blob/6f0735c969c518f9262281ba5539054b14e0d5d4/setup.py#L162

PR #229, opened 2018-11-22
 Removing Multipolygon

VasavanThiru:


docs/source/api_reference/polygon.rst
```diff
@@ -19,6 +22,86 @@ Example
...
+
+
+Example Polygon with hole
+-------------------------
+
+.. code::
+
+  from ipyleaflet import Map, Polygon
+
+  Hpolygon = Polygon(
```
martinRenou: It's unusual to use a capital letter for variable names, we generally use a capital letter for class names. You can just name it `polygon`

docs/source/api_reference/polygon.rst
```diff
@@ -19,6 +22,86 @@ Example
...
+
+
+Example Multipolygon
+--------------------
+
+.. code::
+
+    from ipyleaflet import Map, Polygon
+
+    Multipolygon = Polygon(
```
martinRenou: Same here :) Just `polygon` is fine

PR #273, opened 2019-01-28
Implement LocalTileLayer

### LINKED ISSUE #257: Serve local tiles with notebook
Today it is not possible to set a path to local tiles, with something like:
```python
l = {'url': 'file:///home/user/path/to/tiles/{z}/{x}/{y}.png'}
m.add_layer(l)
```
Instead, the directory of the tiles has to be served with e.g.:
```bash
python -m http.server --bind 127.0.0.1
```
And the URL has to be set to `http://127.0.0.1:8000/{z}/{x}/{y}.png`.
Could it be done transparently with the notebook server?

martinRenou:
Should fix #257

@davidbrochart Would you mind trying this implementation and telling me what you think? :)

### DIFF
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -237,6 +237,13 @@ class TileLayer(RasterLayer):
         self._load_callbacks.register_callback(callback, remove=remove)
 
 
+class LocalTileLayer(TileLayer):
+    _view_name = Unicode('LeafletLocalTileLayerView').tag(sync=True)
+    _model_name = Unicode('LeafletLocalTileLayerModel').tag(sync=True)
+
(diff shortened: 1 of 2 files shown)


davidbrochart:
At first sight it looks like it's prefixing the path with `/notebooks/`, which results in a 404 error.

martinRenou:
Yes, the path is relative to where the notebook is running

davidbrochart:
Actually that doesn't seem to be the issue. What is supposed to be passed to `path` in `m.add_layer(LocalTilesLayer(path='/path/to/tiles/{z}/{x}/{y}.png'))`? An absolute path to the tiles? A path relative to the directory where the notebook was launched? Does it have to be prefixed with `files/`?

PR #281, opened 2019-01-31
Changes for NASA GIBS sample imagery endpoints.

GalacticMatt:
- Updating GIBS URLs to use the offical endpoints and a more recent date.
- Adding VIIRS True Color Corrected Reflectance.
- Updating VIIRS Earth at Night to use the newer layer.  Technically "Black Marble", but still the earth at night.


martinRenou:
Thanks for the PR :)

PR #282, opened 2019-01-31
Cleaning ipyleaflet

VasavanThiru:


js/src/layers/LocalTileLayer.js
```diff
@@ -0,0 +1,40 @@
+var widgets = require('@jupyter-widgets/base');
+var _ = require('underscore');
+var L = require('leaflet');
```
martinRenou: Every time you have
```javascript
var L = require('leaflet');
```
It should be:
```javascript
var L = require('../leaflet.js');
```
Otherwise I think webpack will pick the `leaflet` library directly (without the plugins), not your file.

js/src/layers/FeatureGroup.js
```diff
@@ -0,0 +1,33 @@
+var widgets = require('@jupyter-widgets/base');
+var _ = require('underscore');
+var L = require('leaflet');
+var layergroup = require('./LayerGroup.js')
+
+
+
+// Model
```
martinRenou: Nitpicking here, would you mind removing the extra empty lines that you added? Also, `// Model` and `// View` comments do not add information or readability to the code, so I would remove them too.

VasavanThiru: Yes, I see

VasavanThiru:
Maybe It's better now

js/src/jupyter-leaflet.js
```diff
@@ -1,1755 +1,107 @@
 var widgets = require('@jupyter-widgets/base');
 var _ = require('underscore');
```
martinRenou: Those are not needed anymore in this file

js/src/jupyter-leaflet.js
```diff
@@ -1,1755 +1,107 @@
...
-        });
-    }
-}, {
-    serializers: _.extend({
-        layers : { deserialize: widgets.unpack_models },
-        controls : { deserialize: widgets.unpack_models }
-    }, widgets.DOMWidgetModel.serializers)
-});
+var L = require('./leaflet.js');
+var utils = require('./utils.js');
```
martinRenou: Those `require`s are not needed either

js/src/leaflet.js
```diff
@@ -0,0 +1,25 @@
+
```
martinRenou: Extra empty line at beginning of the file

js/src/map/Map.js
```diff
@@ -0,0 +1,286 @@
+var widgets = require('@jupyter-widgets/base');
```
martinRenou: I would not put `Map.js` alone in a directory, I guess it's fine to put it inside of `js/src/` directory

js/src/utils.js
```diff
@@ -0,0 +1,33 @@
+
```
martinRenou: extra empty line

PR #289, opened 2019-02-12
Removing render function for all controls

VasavanThiru:


js/src/controls/LayersControl.js
```diff
@@ -28,21 +28,17 @@ var LeafletLayersControlView = LeafletControlView.extend({
...
         delete this.obj;
         this.create_obj();
     },
 
     create_obj: function () {
+        var that = this;
+        this.listenTo(this.map_view.model, 'change:layers', function() {
+            that.toggle_obj();
+        });
         var that = this;
```
martinRenou: This line is now duplicated, you can remove this one

js/src/controls/LayersControl.js
```diff
@@ -28,21 +28,17 @@ var LeafletLayersControlView = LeafletControlView.extend({
...
-
     toggle_obj: function () {
         this.obj.remove();
         delete this.obj;
         this.create_obj();
     },
 
     create_obj: function () {
+        var that = this;
+        this.listenTo(this.map_view.model, 'change:layers', function() {
```
martinRenou: This should be in a `model_events` method

js/src/controls/SplitMapControl.js
```diff
@@ -25,21 +25,11 @@ var LeafletSplitMapControlView = LeafletControlView.extend({
         LeafletSplitMapControlView.__super__.initialize.apply(this, arguments);
         this.map_view = this.options.map_view;
     },
-
-    render: function () {
+    
+    create_obj: function () {
         var that = this;
         var left_layer = this.model.get('left_layer');
         var right_layer = this.model.get('right_layer');
```
martinRenou: `that` is not used, you can remove it. Also, `right_layer` and `left_layer` are defined twice in your method, but not with the same definition, I guess those three lines should be removed.

martinRenou:
Great :) Thanks

PR #295, opened 2019-02-19
Load Leaflet in no-conflict mode if L is already defined

drwelby:
Leaflet has a no-conflict mode where if `L` is already defined, it can be restored as `window.L` with Leaflet assigned to another variable of choice.

In this specific case, if another Jupyter widget is already trying to use it's own version of Leaflet, loading ipyleaflet will redefine the existing Leaflet library, and any control or plugins that were previously loaded into the first Leaflet namespace will disappear.

This fix checks to see if `L` is already defined, and if it is, it uses `Leaflet.noConflict()` to restore the previous `L` and moves ipyleaflet's Leaflet to the variable `ipyL`.

Users who only use ipyleaflet will not be impacted as Leaflet will load as `L` in the standard manner.

### DIFF
--- a/js/src/index.js
+++ b/js/src/index.js
@@ -13,5 +13,11 @@ require('leaflet/dist/images/marker-icon.png');
 require('leaflet/dist/images/marker-icon-2x.png');
 
 // Export everything from jupyter-leaflet and the npm package version number.
+hasL = (typeof(window.L) != 'undefined');
 module.exports = require('./jupyter-leaflet.js');
 module.exports['version'] = require('../package.json').version;
+
+if (hasL) {
+    ipyL = L.noConflict();
+}
+
--- a/js/src/index.js
+++ b/js/src/index.js
@@ -18,6 +18,7 @@ module.exports = require('./jupyter-leaflet.js');
 module.exports['version'] = require('../package.json').version;
 
 if (hasL) {
+    console.log("Existing `L` detected, running ipyleaflet's Leaflet in no-conflict mode as `ipyL`");
     ipyL = L.noConflict();
 }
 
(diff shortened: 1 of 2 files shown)


js/src/index.js
```diff
@@ -13,5 +13,12 @@ require('leaflet/dist/images/marker-icon.png');
 require('leaflet/dist/images/marker-icon-2x.png');
 
 // Export everything from jupyter-leaflet and the npm package version number.
+hasL = (typeof(window.L) != 'undefined');
 module.exports = require('./jupyter-leaflet.js');
 module.exports['version'] = require('../package.json').version;
+
+if (hasL) {
+    console.log("Existing `L` detected, running ipyleaflet's Leaflet in no-conflict mode as `ipyL`");
+    ipyL = L.noConflict();
```
martinRenou: I'm not entirely sure I understand. Shouldn't we use `ipyL` everywhere in ipyleaflet in this case? Instead of `L`? I mean, what do we do with this variable?

drwelby: The ipyleaflet code will still use L since when it is all loaded L is the internal Leaflet namespace within the module.

Leaflet's no-conflict storage of the existing global L is independent of any module exports. So that's why we first check if there will be a conflict, then load ipyleaflet, then restore the original global L.

I have only tested using the ipyleaflet functions we use. Without a test suite it might be better to pull this change initially to its own branch for others to check.

martinRenou: So, if I understand correctly, if you want to restore the global `L`, shouldn't this line of code be: `L = L.noConflict();` or `window.L = L.noConflict();`?

drwelby: L. noConflict() restores the previous global L to `window.L`, and returns the current Leaflet library so you can assign it to a different variable name:

https://leafletjs.com/reference-1.4.0.html#noconflict

martinRenou: Ok perfect then :) Thanks!

drwelby: Thanks Martin!

PR #306, opened 2019-03-12
Widget control

### LINKED ISSUE #288: Custom controls
Is it possible to add custom widgets (ie, any ipywidget) as a control ?
I see I can add an ipywidget in a Popup, but adding it as a control would open nice possibilities, like color bars, legends...

VasavanThiru:
widgetcontrol

Fix #288

### DIFF
--- a/js/src/controls/WidgetControl.js
+++ b/js/src/controls/WidgetControl.js
@@ -8,16 +8,11 @@ var LeafletControlModel = control.LeafletControlModel;
 L.Control.WidgetControl = L.Control.extend({
     options: {
         position: 'topleft',
-        widget : 'null',
     },
 
     _updateContent: function () {
         if (!this._content) { return; }
-        var node = this._container
-        L.DomEvent.disableClickPropagation(node);
-        L.DomEvent.disableScrollPropagation(node);
-        var content = this._content;
-        node.appendChild(content);
+        this._container.appendChild(this._content);
     },
 
     update: function () {
@@ -36,9 +31,10 @@ L.Control.WidgetControl = L.Control.extend({
     },
 
     onAdd: function (map) {
-        var container = L.DomUtil.create('div', 'leaflet-widgetcontrol');
-        this._container = container ;
-        return container;
+        this._container = L.DomUtil.create('div', 'leaflet-widgetcontrol');
+        L.DomEvent.disableClickPropagation(this._container);
+        L.DomEvent.disableScrollPropagation(this._container);
+        return this._container;
     },
 });
 
@@ -67,12 +63,13 @@ var LeafletWidgetControlView = LeafletControlView.extend({
         this.widget_view = undefined;
     },
 
-    set_widget: function(value){
+    set_widget: function(widget_model){
         if (this.widget_view){
             this.widget_view.remove();
+            this.widget_view = undefined;
         }
-        if (value){
-            return this.create_child_view(value).then((view)=>{
+        if (widget_model){
+            return this.create_child_view(widget_model).then((view)=>{
                 this.widget_view = view;
                 this.obj.setContent(view.el)
             })
--- a/js/src/controls/WidgetControl.js
+++ b/js/src/controls/WidgetControl.js
@@ -8,7 +8,7 @@ var LeafletControlModel = control.LeafletControlModel;
 
 L.Control.WidgetControl = L.Control.extend({
     options: {
-        position: 'topleft',
+        position: 'topright',
         title: {
             'false': 'View WidgetControl',
             'true': 'Exit WidgetControl'
@@ -61,6 +61,7 @@ L.Control.WidgetControl = L.Control.extend({
 		return this;
     },
     */
+
     _updateContent: function () {
 
 		if (!this._content) { return; }
@@ -74,6 +75,9 @@ L.Control.WidgetControl = L.Control.extend({
 
 		//var content = (typeof this._content === 'function') ? this._content(this._source || this) : this._content;
         var node = this._container
+        L.DomEvent.disableClickPropagation(node);
+        L.DomEvent.disableScrollPropagation(node);
+
         var content = this._content;
 		while (node.hasChildNodes()) {
 			node.removeChild(node.firstChild);
@@ -95,7 +99,7 @@ L.Control.WidgetControl = L.Control.extend({
 		//this._container.style.visibility = '';
 
 		//this._adjustPan();
-},
+    },
 
 
     getContent: function(){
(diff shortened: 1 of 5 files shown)


js/src/controls/WidgetControl.js
```diff
@@ -0,0 +1,83 @@
+var widgets = require('@jupyter-widgets/base');
+var _ = require('underscore');
+var L = require('../leaflet.js');
+var control = require('./Control.js');
+var LeafletControlView = control.LeafletControlView;
+var LeafletControlModel = control.LeafletControlModel;
+
+L.Control.WidgetControl = L.Control.extend({
+    options: {
+        position: 'topright',
```
martinRenou: The default value is 'topleft' at other places, maybe change this one to 'topleft' too?

js/src/controls/WidgetControl.js
```diff
@@ -0,0 +1,83 @@
...
+    },
+
+    _updateContent: function () {
+		if (!this._content) { return; }
```
martinRenou: There is a mix of tabs and spaces

js/src/controls/WidgetControl.js
```diff
@@ -0,0 +1,83 @@
...
+        var node = this._container
+        L.DomEvent.disableClickPropagation(node);
+        L.DomEvent.disableScrollPropagation(node);
+        var content = this._content;
+		while (node.hasChildNodes()) {
+			node.removeChild(node.firstChild);
+		}
```
martinRenou: That is not a clean way to remove old widget views. You need to call `view.remove()` on the old view if it exists.

js/src/controls/WidgetControl.js
```diff
@@ -0,0 +1,83 @@
...
+
+var LeafletWidgetControlView = LeafletControlView.extend({
+    initialize: function (parameters) {
+        LeafletWidgetControlView.__super__.initialize.apply(this, arguments);
+        this.map_view = this.options.map_view;
+    },
+
+    create_obj: function () {
+        this.obj = L.control.widgetcontrol(this.get_options());
+        this.create_child_view(this.model.get('widget')).then((view)=>{
```
martinRenou: Maybe you should return the promise?

SylvainCorlay:
Would be nice to have a screenshot in the PR!

js/src/controls/WidgetControl.js
```diff
@@ -0,0 +1,94 @@
...
+var _ = require('underscore');
+var L = require('../leaflet.js');
+var control = require('./Control.js');
+var LeafletControlView = control.LeafletControlView;
+var LeafletControlModel = control.LeafletControlModel;
+
+L.Control.WidgetControl = L.Control.extend({
+    options: {
+        position: 'topleft',
+        widget : 'null',
```
martinRenou: As far as I can see you never use this `widget` option, I guess you can remove it

js/src/controls/WidgetControl.js
```diff
@@ -0,0 +1,94 @@
...
+    },
+
+    _updateContent: function () {
+        if (!this._content) { return; }
+        var node = this._container
+        L.DomEvent.disableClickPropagation(node);
+        L.DomEvent.disableScrollPropagation(node);
```
martinRenou: Maybe you can move those two lines:
```javascript
L.DomEvent.disableClickPropagation(node);
L.DomEvent.disableScrollPropagation(node);
```
in the `onAdd` method? After `this._container` is created? So that you call it only once.

js/src/controls/WidgetControl.js
```diff
@@ -0,0 +1,94 @@
...
+        var content = this._content;
+        node.appendChild(content);
```
martinRenou: No need to do:
```javascript
var node = this._container;
var content = this._content;
node.appendChild(content);
```
You can directly do:
```javascript
this._container.appendChild(this._content);
```

js/src/controls/WidgetControl.js
```diff
@@ -0,0 +1,94 @@
...
+    setContent: function(content){
+        this._content = content;
+        this.update();
+        return this;
+    },
+
+    onAdd: function (map) {
+        var container = L.DomUtil.create('div', 'leaflet-widgetcontrol');
+        this._container = container ;
+        return container;
```
martinRenou: You can also save one line here ;)
```javascript
this._container = L.DomUtil.create('div', 'leaflet-widgetcontrol');
return this._container;
```
no need for a `container` variable

js/src/controls/WidgetControl.js
```diff
@@ -0,0 +1,94 @@
...
+var LeafletWidgetControlView = LeafletControlView.extend({
+    initialize: function (parameters) {
+        LeafletWidgetControlView.__super__.initialize.apply(this, arguments);
+        this.map_view = this.options.map_view;
+        this.widget_view = undefined;
+    },
+
+    set_widget: function(value){
+        if (this.widget_view){
+            this.widget_view.remove();
```
martinRenou: Maybe reset `this.widget_view` to undefined after the remove?

js/src/controls/WidgetControl.js
```diff
@@ -0,0 +1,94 @@
...
+});
+
+var LeafletWidgetControlView = LeafletControlView.extend({
+    initialize: function (parameters) {
+        LeafletWidgetControlView.__super__.initialize.apply(this, arguments);
+        this.map_view = this.options.map_view;
+        this.widget_view = undefined;
+    },
+
+    set_widget: function(value){
```
martinRenou: That's really nitpicking, but it would be clearer if you rename `value` to `widget_model`. So you have `widget_view` and `widget_model`

VasavanThiru: Thanks a lot for all your remarks !  @martinRenou

js/src/controls/WidgetControl.js
```diff
@@ -0,0 +1,91 @@
...
+
+    getContent: function(){
+        return this._content;
+    },
+
+    setContent: function(content){
+        this._content = content;
+        this.update();
+        return this;
+    },
```
martinRenou: One last thing ;)
You've got two unneeded methods `_updateContent` and `update`. Technically your code results in this:
```javascript
setContent: function(content){
    if (!this._map) { return; }
    this._content = content;
    this._container.appendChild(this._content);
    return this;
},
```

js/src/controls/WidgetControl.js
martinRenou: Sure ;)

martinRenou:
I know how to fix the use case of adding a map inside of the widget control. I'll open a separate PR with the fix :)

PR #310, opened 2019-03-15
Allow maps in WidgetControl

martinRenou:
widget_control

@VasavanThiru


VasavanThiru:
Yes, it's very nice !!

PR #328, opened 2019-04-07
Bump jupyterlab-manager to 0.38

jtpio:
Binder images currently use JupyterLab 0.35 by default.
The compatible version for `@jupyter-widgets/jupyterlab-manager` is 0.38.


martinRenou:
Is it still the case as of today @jtpio ? Should we update this requirement?

Asking because of #501

jtpio:
Ah right, we should be able to remove the pin now.

PR #330, opened 2019-04-08
Use JLab background color for widget controls

jtpio:
A small / simple change to fix the following glitch when using the JupyterLab Dark theme:

### Before

image

### After

image

In the context of the classic notebook, the background will default to the original `#ffffff` value.

### DIFF
--- a/js/src/jupyter-leaflet.css
+++ b/js/src/jupyter-leaflet.css
@@ -5,7 +5,7 @@
 }
 
 .leaflet-widgetcontrol {
-    background-color: #ffffff;
+    background: var(--jp-layout-color1);
     border-radius: 4.5px;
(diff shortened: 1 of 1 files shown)


wolfv:
Cool! Shouldn't the round knob also have a lighter color then?

jtpio:
> Cool! Shouldn't the round knob also have a lighter color then?

It looks similar to the default IntSlider in Dark Theme:

image

> **EDIT:** Even the base zoom controls should support JLab themes

For now it's still ok because they are quite readable even though they don't match with the theme. But yes ideally they should.

martinRenou:
Thanks again :)

jtpio:
:+1:

PR #334, opened 2019-04-15
DOC: Add crs option on Map

### LINKED ISSUE #327: Document the support for "simple" CRS's
Perhaps I missed it, but I could not find any way to create an `ipyleaflet` `Map` that uses the built-in leaflet CRS's.  In particular I'm interested in the "simple"  one (E.g., https://leafletjs.com/examples/crs-simple/crs-simple.html). The idea here is that applying leaflet to more generic datasets than geo maps (my specific interest is a variety of astronomy applications) is a lot easier with the `Simple` CRS.

Depending on the answer, this issue could be either  "feature request", "where would I look to access the JS code that initializes map so that I can make a PR", or "the explanation of how should be in the docs".

Note this is a bit distinct from #70 mainly because that's a much more general thing.  This is about the built-in leaflet CRS's (especially `simple`)

pllim:
Fix #327.

cc @eteq

### DIFF
--- a/docs/source/api_reference/map.rst
+++ b/docs/source/api_reference/map.rst
@@ -41,6 +41,7 @@ center                      (0.0, 0.0)                          Initial geograph
 zoom                        12                                  Initial map zoom level
 max_zoom                    18
 min_zoom                    1
+crs                         'EPSG3857'                          Coordinate reference system, which can be 'Earth', 'EPSG3395', 'EPSG3857', 'EPSG4326', 'Base', or 'Simple'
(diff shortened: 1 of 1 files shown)


martinRenou:
Awesome! thanks :)

eteq:
Oh wow. That was very simple - guess should have done it myself weeks ago, but thanks @pllim !

martinRenou:
We need more contributors to ipyleaflet ;) PRs are really appreciated! I don't really have time to give love to ipyleaflet, even if I would like to.

Issue #167, opened 2018-04-25
Change cursor style over Map


fitoprincipe:
It would be great if somehow the map's cursor could be changed to any of the supported styles. It is not hard to do it on the JavaScript side, just changing Map's `cursor` style ([stackoverflow](https://stackoverflow.com/questions/37518278/leaflet-how-can-i-change-the-grabbing-hand-cursor/41232864?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa)), but I wouldn't know how to do it on ipyleaflet widget

SylvainCorlay:
OK, I think that we need to have a `style` widget on the map just like on the ipywidgets slider to set CSS properties of the map.

fitoprincipe:
That would be great! thank you

fitoprincipe:
Hi @martinRenou. Any chance to include this in the Project's TODO?

martinRenou:
Sure. But it doesn't mean I'll fix it soon... I added "simple" issues to the TODO list, issues that are newcomers friendly.

fitoprincipe:
I understand, thank you @martinRenou, I wish I'd have some time to learn how all works (specially the JS side) to fix it myself. I've been reading but I'm not there yet. =)

fitoprincipe:
I need some guide (help) here..

In  `leaflet.py` I made 2 changes:

1. A new class:

``` python
class MapStyle(Style, CoreWidget, Widget):
    """ Map Style Widget """
    _model_name = Unicode('MapStyleModel').tag(sync=True)
    cursor = Unicode(help='Cursor type').tag(sync=True)
```
2. imported `InstanceDict` from `ipywidgets.widgets.trait_types` and add an attribute in `Map`: 

``` python
style = InstanceDict(MapStyle).tag(sync=True, **widget_serialization)
```

In `Map.js` I created a new class:

``` javascript
var MapStyleModel = widgets.WidgetModel.extend({
    defaults: _.extend({}, widgets.StyleModel.prototype.defaults, {
        _model_name : "MapStyleModel",
        _model_module : "jupyter-leaflet",
        }),
 
    styleProperties = {
        cursor: {
            selector: '.leaflet-container',
            attribute: 'cursor',
            default: 'move'
        }
    }
}); 
```
I added this new class in `exports`:

``` javascript
module.exports = {
  LeafletMapView : LeafletMapView,
  LeafletMapModel : LeafletMapModel,
[...]

PR #339, opened 2019-04-24
add MapStyle to change Map's css style, aiming to change the cursor type

fitoprincipe:
Trying to solve issue https://github.com/jupyter-widgets/ipyleaflet/issues/167

### DIFF
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -809,7 +809,8 @@ class Map(DOMWidget, InteractMixin):
 
     options = List(trait=Unicode).tag(sync=True)
 
-    style = InstanceDict(MapStyle).tag(sync=True, **widget_serialization)
+    default_style = InstanceDict(MapStyle).tag(sync=True, **widget_serialization)
+    dragging_style = InstanceDict(MapStyle, default_value={'cursor': 'move'}).tag(sync=True, **widget_serialization)
 
     @default('options')
     def _default_options(self):
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -753,7 +753,7 @@ class MapStyle(Style, Widget):
     """ Map Style Widget """
     _model_name = Unicode('LeafletMapStyleModel').tag(sync=True)
     _model_module = Unicode("jupyter-leaflet").tag(sync=True)
-    cursor = Unicode('hand', help='Cursor type').tag(sync=True)
+    cursor = Unicode('grab', help='Cursor type').tag(sync=True)
 
 
 class Map(DOMWidget, InteractMixin):
@@ -811,7 +811,7 @@ class Map(DOMWidget, InteractMixin):
 
     style = InstanceDict(MapStyle).tag(sync=True, **widget_serialization)
     dragging_style = Dict({'cursor':'move'}).tag(sync=True)
-    default_style = Dict({'cursor':'hand'}).tag(sync=True)
+    default_style = Dict({'cursor':'grab'}).tag(sync=True)
 
     @default('options')
     def _default_options(self):
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -2,7 +2,7 @@ import copy
 
 from ipywidgets import (
     Widget, DOMWidget, Box, Color, CallbackDispatcher, widget_serialization,
-    interactive, Style, CoreWidget
+    interactive, Style
 )
 
 from ipywidgets.widgets.trait_types import InstanceDict
@@ -809,8 +809,9 @@ class Map(DOMWidget, InteractMixin):
 
     options = List(trait=Unicode).tag(sync=True)
 
-    default_style = InstanceDict(MapStyle).tag(sync=True, **widget_serialization)
-    dragging_style = InstanceDict(MapStyle, default_value={'cursor': 'move'}).tag(sync=True, **widget_serialization)
+    style = InstanceDict(MapStyle).tag(sync=True, **widget_serialization)
+    dragging_style = Dict({'cursor':'move'}).tag(sync=True)
+    default_style = Dict({'cursor':'hand'}).tag(sync=True)
 
     @default('options')
     def _default_options(self):
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -753,7 +753,7 @@ class MapStyle(Style, Widget):
     """ Map Style Widget """
     _model_name = Unicode('LeafletMapStyleModel').tag(sync=True)
     _model_module = Unicode("jupyter-leaflet").tag(sync=True)
-    cursor = Unicode(help='Cursor type').tag(sync=True)
+    cursor = Unicode('move', help='Cursor type').tag(sync=True)
 
 
 class Map(DOMWidget, InteractMixin):
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -2,8 +2,10 @@ import copy
 
 from ipywidgets import (
     Widget, DOMWidget, Box, Color, CallbackDispatcher, widget_serialization,
-    interactive
+    interactive, Style, CoreWidget
 )
+
+from ipywidgets.widgets.trait_types import InstanceDict
 
 from traitlets import (
     Float, Unicode, Int, Tuple, List, Instance, Bool, Dict, Enum,
@@ -747,6 +749,13 @@ class DrawControl(Control):
         self.send({'msg': 'clear_markers'})
 
 
+class MapStyle(Style, Widget):
+    """ Map Style Widget """
+    _model_name = Unicode('LeafletMapStyleModel').tag(sync=True)
+    _model_module = Unicode("jupyter-leaflet").tag(sync=True)
+    cursor = Unicode(help='Cursor type').tag(sync=True)
+
+
 class Map(DOMWidget, InteractMixin):
     _view_name = Unicode('LeafletMapView').tag(sync=True)
     _model_name = Unicode('LeafletMapModel').tag(sync=True)
@@ -799,6 +808,8 @@ class Map(DOMWidget, InteractMixin):
     # marker_zoom_animation = Bool(?).tag(sync=True, o=True)
 
     options = List(trait=Unicode).tag(sync=True)
+
+    style = InstanceDict(MapStyle).tag(sync=True, **widget_serialization)
 
     @default('options')
     def _default_options(self):
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -753,7 +753,7 @@ class MapStyle(Style, Widget):
     """ Map Style Widget """
     _model_name = Unicode('LeafletMapStyleModel').tag(sync=True)
     _model_module = Unicode("jupyter-leaflet").tag(sync=True)
-    cursor = Unicode('move', help='Cursor type').tag(sync=True)
+    cursor = Unicode('hand', help='Cursor type').tag(sync=True)
 
 
 class Map(DOMWidget, InteractMixin):
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -24,6 +24,12 @@ from ._version import EXTENSION_VERSION
 
 def_loc = [0.0, 0.0]
 allowed_crs = ['Earth', 'EPSG3395', 'EPSG3857', 'EPSG4326', 'Base', 'Simple']
+allowed_cursor = ['alias', 'cell', 'grab', 'move', 'crosshair', 'context-menu',
+                  'n-resize', 'ne-resize', 'e-resize', 'se-resize', 's-resize',
+                  'sw-resize', 'w-resize', 'nw-resize', 'nesw-resize',
+                  'nwse-resize', 'row-resize', 'col-resize', 'copy', 'default',
+                  'grabbing', 'help', 'no-drop', 'not-allowed', 'pointer',
+                  'progress', 'text', 'wait', 'zoom-in', 'zoom-out']
 
 
 def basemap_to_tiles(bm, day='yesterday', **kwargs):
@@ -753,7 +759,7 @@ class MapStyle(Style, Widget):
     """ Map Style Widget """
     _model_name = Unicode('LeafletMapStyleModel').tag(sync=True)
     _model_module = Unicode("jupyter-leaflet").tag(sync=True)
-    cursor = Unicode('grab', help='Cursor type').tag(sync=True)
+    cursor = Enum(values=allowed_cursor, default_value='grab').tag(sync=True)
 
 
 class Map(DOMWidget, InteractMixin):
@@ -810,8 +816,12 @@ class Map(DOMWidget, InteractMixin):
     options = List(trait=Unicode).tag(sync=True)
 
     style = InstanceDict(MapStyle).tag(sync=True, **widget_serialization)
-    dragging_style = Dict({'cursor':'move'}).tag(sync=True)
-    default_style = Dict({'cursor':'grab'}).tag(sync=True)
+    default_style = InstanceDict(MapStyle).tag(sync=True, **widget_serialization)
+    dragging_style = InstanceDict(MapStyle).tag(sync=True, **widget_serialization)
+
+    @default('dragging_style')
+    def _default_dragging_style(self):
+        return {'cursor': 'move'}
 
     @default('options')
     def _default_options(self):
--- a/js/src/Map.js
+++ b/js/src/Map.js
@@ -10,16 +10,16 @@ var LeafletMapStyleModel = widgets.StyleModel.extend({
     defaults: _.extend({}, widgets.StyleModel.prototype.defaults, {
         _model_name : "LeafletMapStyleModel",
         _model_module : "jupyter-leaflet",
-        }),
-
-    styleProperties: {
-        cursor: {
-            selector: '.leaflet-grab',
-            attribute: 'cursor',
-            default: 'move'
-        }
-    },
+    })
 });
+
+LeafletMapStyleModel.styleProperties = {
+     cursor: {
+         selector: '',
+         attribute: 'cursor',
+         default: 'hand'
+     }
+};
 
 var LeafletMapModel = widgets.DOMWidgetModel.extend({
     defaults: _.extend({}, widgets.DOMWidgetModel.prototype.defaults, {
--- a/js/src/Map.js
+++ b/js/src/Map.js
@@ -6,6 +6,21 @@ var LeafletWidgetView = utils.LeafletWidgetView;
 var LeafletDOMWidgetView = utils.LeafletDOMWidgetView;
 var def_loc = [0.0, 0.0];
 
+var LeafletMapStyleModel = widgets.StyleModel.extend({
+    defaults: _.extend({}, widgets.StyleModel.prototype.defaults, {
+        _model_name : "LeafletMapStyleModel",
+        _model_module : "jupyter-leaflet",
+        }),
+
+    styleProperties: {
+        cursor: {
+            selector: '.leaflet-grab',
+            attribute: 'cursor',
+            default: 'move'
+        }
+    },
+});
+
 var LeafletMapModel = widgets.DOMWidgetModel.extend({
     defaults: _.extend({}, widgets.DOMWidgetModel.prototype.defaults, {
         _view_name : "LeafletMapView",
@@ -54,7 +69,8 @@ var LeafletMapModel = widgets.DOMWidgetModel.extend({
         options : [],
         layers : [],
         controls : [],
-        crs: 'EPSG3857'
+        crs: 'EPSG3857',
+        style: 'move'
     }),
 
     update_bounds: function() {
@@ -284,4 +300,5 @@ var LeafletMapView = utils.LeafletDOMWidgetView.extend({
 module.exports = {
   LeafletMapView : LeafletMapView,
   LeafletMapModel : LeafletMapModel,
+  LeafletMapStyleModel: LeafletMapStyleModel,
 };
--- a/js/src/Map.js
+++ b/js/src/Map.js
@@ -73,7 +73,17 @@ var LeafletMapModel = widgets.DOMWidgetModel.extend({
         style: null,
         default_style: {'cursor': 'grab'},
         dragging_style: {'cursor': 'move'},
+        _dragging: false
     }),
+
+    update_style: function() {
+        if (!this.get('_dragging')) {
+            var new_style = this.get('default_style');
+        } else {
+            var new_style = this.get('dragging_style');
+        };
+        this.get('style').set('cursor', new_style.cursor);
+    },
 
     update_bounds: function() {
         var that = this;
@@ -198,14 +208,16 @@ var LeafletMapView = utils.LeafletDOMWidgetView.extend({
                 that.model.set('center', [c.lat, c.lng]);
                 that.dirty = false;
             };
-            that.model.get('style').set('cursor', that.model.get('default_style').cursor)
             that.model.update_bounds().then(function() {
                 that.touch();
             });
+            that.model.set('_dragging', false);
+            that.model.update_style();
         });
 
         this.obj.on('movestart', function (e) {
-            that.model.get('style').set('cursor', that.model.get('dragging_style').cursor)
+            that.model.set('_dragging', true);
+            that.model.update_style();
         });
 
         this.obj.on('zoomend', function (e) {
@@ -268,6 +280,12 @@ var LeafletMapView = utils.LeafletDOMWidgetView.extend({
             this.model.update_bounds().then(function() {
                 that.touch();
             });
+        }, this);
+        this.listenTo(this.model, 'change:dragging_style', function () {
+            this.model.update_style();
+        }, this);
+        this.listenTo(this.model, 'change:default_style', function () {
+            this.model.update_style();
         }, this);
     },
 
--- a/js/src/Map.js
+++ b/js/src/Map.js
@@ -70,7 +70,7 @@ var LeafletMapModel = widgets.DOMWidgetModel.extend({
         layers : [],
         controls : [],
         crs: 'EPSG3857',
-        style: 'move'
+        style: null
     }),
 
     update_bounds: function() {
@@ -99,7 +99,8 @@ var LeafletMapModel = widgets.DOMWidgetModel.extend({
 }, {
     serializers: _.extend({
         layers : { deserialize: widgets.unpack_models },
-        controls : { deserialize: widgets.unpack_models }
+        controls : { deserialize: widgets.unpack_models },
+        style : { deserialize: widgets.unpack_models }
     }, widgets.DOMWidgetModel.serializers)
 });
 
--- a/js/src/Map.js
+++ b/js/src/Map.js
@@ -70,8 +70,9 @@ var LeafletMapModel = widgets.DOMWidgetModel.extend({
         layers : [],
         controls : [],
         crs: 'EPSG3857',
-        default_style: null,
-        dragging_style: null,
+        style: null,
+        default_style: {'cursor': 'hand'},
+        dragging_style: {'cursor': 'move'},
     }),
 
     update_bounds: function() {
@@ -101,8 +102,7 @@ var LeafletMapModel = widgets.DOMWidgetModel.extend({
     serializers: _.extend({
         layers : { deserialize: widgets.unpack_models },
         controls : { deserialize: widgets.unpack_models },
-        default_style : { deserialize: widgets.unpack_models },
-        dragging_style : { deserialize: widgets.unpack_models },
+        style: { deserialize: widgets.unpack_models },
     }, widgets.DOMWidgetModel.serializers)
 });
 
@@ -198,28 +198,14 @@ var LeafletMapView = utils.LeafletDOMWidgetView.extend({
                 that.model.set('center', [c.lat, c.lng]);
                 that.dirty = false;
             };
-            var style_view = Object.keys(that.model.get('default_style').views)[0];
-            that.model.get('default_style').views[style_view].then(function(view) {
-                view.style();
-            });
-            var dragging_style_view = Object.keys(that.model.get('dragging_style').views)[0];
-            that.model.get('dragging_style').views[dragging_style_view].then(function(view) {
-                view.unstyle();
-            });
+            that.model.get('style').set('cursor', that.model.get('default_style').cursor)
             that.model.update_bounds().then(function() {
                 that.touch();
             });
         });
 
         this.obj.on('movestart', function (e) {
-            var style_view = Object.keys(that.model.get('default_style').views)[0];
-            that.model.get('default_style').views[style_view].then(function(view) {
-                view.unstyle();
-            });
-            var dragging_style_view = Object.keys(that.model.get('dragging_style').views)[0];
-            that.model.get('dragging_style').views[dragging_style_view].then(function(view) {
-                view.style();
-            });
+            that.model.get('style').set('cursor', that.model.get('dragging_style').cursor)
         });
 
         this.obj.on('zoomend', function (e) {
--- a/js/src/Map.js
+++ b/js/src/Map.js
@@ -71,8 +71,8 @@ var LeafletMapModel = widgets.DOMWidgetModel.extend({
         controls : [],
         crs: 'EPSG3857',
         style: null,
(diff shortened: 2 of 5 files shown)


ipyleaflet/leaflet.py
```diff
@@ -747,6 +749,13 @@ def clear_markers(self):
         self.send({'msg': 'clear_markers'})
 
 
+class MapStyle(Style, Widget):
+    """ Map Style Widget """
+    _model_name = Unicode('LeafletMapStyleModel').tag(sync=True)
+    _model_module = Unicode("jupyter-leaflet").tag(sync=True)
+    cursor = Unicode(help='Cursor type').tag(sync=True)
```
martinRenou: Maybe add the default `'move'` value here? At some point, we could use an `Enum` for allowing only a range of values, but you can leave it as it is for now :)

js/src/Map.js
```diff
@@ -54,7 +69,8 @@ var LeafletMapModel = widgets.DOMWidgetModel.extend({
         options : [],
         layers : [],
         controls : [],
-        crs: 'EPSG3857'
+        crs: 'EPSG3857',
+        style: 'move'
```
martinRenou: Here style must be an instance of `LeafletMapStyleModel`, maybe put `style: null` for now.

js/src/Map.js
```diff
@@ -54,7 +69,8 @@ var LeafletMapModel = widgets.DOMWidgetModel.extend({
...
     }),
```
martinRenou: I think you need to add a new deserializer for the `style` property around line 101

ipyleaflet/leaflet.py
fitoprincipe: I tried that, but as nothing changed I left it as is..

js/src/Map.js
fitoprincipe: I am not sure what deserializers are and what they are for... is this ok?
``` javascript
serializers: _.extend({
        layers : { deserialize: widgets.unpack_models },
        controls : { deserialize: widgets.unpack_models },
        style: { deserialize: widgets.unpack_models }
    }, widgets.DOMWidgetModel.serializers)
```

martinRenou: This should be ok :) serialization is for turning object state (whether it is a Python object or a JavaScript object) into a representation that can be stored or sent to through messaging logic. For instance, here we use JSON representations: we turn Python widget state into JSON representation so that it can be sent from the kernel to the client and vice versa. Deserialization is the opposite operation.
Here you need to tell ipywidgets that style is a widget, so it needs to use a deserializer specific to widgets for this to work :)

martinRenou: Here I'm talking about object states because I have widgets in mind. But you can have serializers/deserializer for arrays for instance. You could represent NumPy arrays as a JSON object, and send it to the front-end which will create its own JavaScript object representing the array, given the JSON object.

fitoprincipe: Ok, I think I get the idea.. Is that why we call `.tag(sync=True, **widget_serialization)` in the python side?

martinRenou: Yep :) You need that for widget attributes.

martinRenou: Does it work now?

fitoprincipe: unfortunately.. nop =(

fitoprincipe: If I do this:
``` javascript
b = Button()
b
```
and then
``` javascript
b.style = {'button_color': 'blue'}
```
it works fine...
but if I do:
``` javascript
m = ipyleaflet.Map()
m
```
and then
``` javascript
m.style = {'cursor': 'crosshair'}
```
it doesn't =(

js/src/Map.js
```diff
@@ -6,6 +6,21 @@ var LeafletWidgetView = utils.LeafletWidgetView;
 var LeafletDOMWidgetView = utils.LeafletDOMWidgetView;
 var def_loc = [0.0, 0.0];
 
+var LeafletMapStyleModel = widgets.StyleModel.extend({
+    defaults: _.extend({}, widgets.StyleModel.prototype.defaults, {
+        _model_name : "LeafletMapStyleModel",
+        _model_module : "jupyter-leaflet",
+        }),
+
+    styleProperties: {
```
martinRenou: I'm wondering if this is the right syntax

fitoprincipe: This is how it looks in `ipywidgets`:

``` typescript
export
class ButtonStyleModel extends StyleModel {
    defaults() {
        return _.extend(super.defaults(), {
            _model_name: 'ButtonStyleModel',
            _model_module: '@jupyter-widgets/controls',
            _model_module_version: JUPYTER_CONTROLS_VERSION,

        });
    }

    public static styleProperties = {
        button_color: {
            selector: '',
            attribute: 'background-color',
            default: null
        },
        font_weight: {
            selector: '',
            attribute: 'font-weight',
            default: ''
        }
    };
}
```

martinRenou:
@SylvainCorlay Do you mind checking this PR? I'm not able to find why it does not work as it is

js/src/Map.js
fitoprincipe: I'm out of the office now.. I'll get back in about 15 hours.. but, just before leaving the office I tried with the compiled version of `ButtonStyleModel` I found in `webpack:///packages/controls/lib/widget_button.js`. I replaced the `LeafletMapStyleModel` in the js side with this:

``` javascript
var __extends = (this && this.__extends) || (function () {
    var extendStatics = Object.setPrototypeOf ||
        ({ __proto__: [] } instanceof Array && function (d, b) { d.__proto__ = b; }) ||
        function (d, b) { for (var p in b) if (b.hasOwnProperty(p)) d[p] = b[p]; };
    return function (d, b) {
        extendStatics(d, b);
        function __() { this.constructor = d; }
        d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
    };
})();
var LeafletMapStyleModel = /** @class */ (function (_super) {
    __extends(LeafletMapStyleModel, _super);
    function LeafletMapStyleModel() {
        return _super !== null && _super.apply(this, arguments) || this;
    }
    LeafletMapStyleModel.prototype.defaults = function () {
        return _.extend(_super.prototype.defaults.call(this), {
            _model_name: 'LeafletMapStyleModel',
        });
    };
    LeafletMapStyleModel.styleProperties = {
        cursor: {
            selector: '',
            attribute: 'cursor',
            default: 'move'
        }
    };
    return LeafletMapStyleModel;
}(widgets.StyleModel));
```
and it worked! =)
But I cannot see what's the difference that makes this work and the other not.. :thinking:

martinRenou: Yes.. I guess you can just do:
```javascript
    LeafletMapStyleModel.styleProperties = {
        cursor: {
            selector: '',
            attribute: 'cursor',
            default: 'move'
        }
    };
```
Just after the `LeafletMapStyleModel` definition. That should work. At some point I would like ipyleaflet to be written in TypeScript.

fitoprincipe: I am testing now.. and have seen that it works when `selector: ''`, but if I try to change another selector style it doesn't work, for example `selector: '.leaflet-grab'`. The thing is that with the option that works, the cursor stays the same in the dragging process, and to return that behavior back I have to set cursor's style to `''`. Maybe `Style` widget cannot change css class properties. Do you think is still good?

martinRenou: I don't know how it works honestly. I'm not sure I get why it works with `selector: ''` and not with `selector: '.leaflet-grab'`. I don't think the `Style` widget changes the css class properties, only the element style attributes. There is also the `.leaflet-dragging` class that you can use as a selector for when the user is dragging the `Map`, maybe that could be useful.

martinRenou: Ok, I think I know why it does not work with the selector. That's because the selector is used here https://github.com/jupyter-widgets/ipywidgets/blob/1f93ad39802d3ee4a6b88bc13064f93f8f796374/packages/base/src/widget_style.ts#L72. The queryselector is used directly from the `el` attribute (the map itself). Just keep it without the selector we can fix it later.

fitoprincipe: yes, but this approach cannot change any class selector, like `.leaflet-grab`, `.leaflet-container` or `.leaflet-dragging`... to me it'd be still very useful to change the cursor this way.. but I don't know if everyone agrees.
I leave some gifs that explain better:

## Changing cursor's type using a css class attribute
drag

## Changing cursor's type using the element css attribute (this approach)
nodrag

martinRenou: I think a way for fixing this later will be to put the `Map` inside of a subcontainer under the `el` element.

martinRenou: Ok, do you mind me pushing to your branch?

fitoprincipe: Not at all! I'm so excited I can contribute to this great project =)

martinRenou: I opened a PR on your fork https://github.com/fitoprincipe/ipyleaflet/pull/1 :) It should fix your selector issue if you merge it on your branch :)

fitoprincipe: Mmm.. works as before.. changes the element's style.. not `.leaflet-grab` style.. and when dragging, the cursor stays the same :thinking:

martinRenou: It cannot change the `.leaflet-grab` class. AFAIK you cannot achieve that in JavaScript. The only thing you can do is select an element given a condition (in our case it needs the `.leaflet-grab` class), then change the element style.

I think what you want to do now is having another style attribute, something like:
```python
class MapStyle(Style, Widget):
    """ Map Style Widget """
    _model_name = Unicode('LeafletMapStyleModel').tag(sync=True)
    _model_module = Unicode("jupyter-leaflet").tag(sync=True)
    
    default_cursor = Unicode('hand', help='Cursor type').tag(sync=True)
    dragging_cursor = Unicode('move', help='Cursor type').tag(sync=True)
```
And on the JS side something like:
```javascript
LeafletMapStyleModel.styleProperties = {
     default_cursor: {
         selector: '.leaflet-grab',
         attribute: 'cursor',
         default: 'hand'
     },
    dragging_cursor: {
         selector: '.leaflet-dragging',
         attribute: 'cursor',
         default: 'move'
     }
};
```
I guess (hope) it will work

fitoprincipe: but, how will it know that it must be applied when dragging if has no access to `.leaflet-dragging` class?? I'll try though =)

martinRenou: I am not sure I understand your question. The `.leaflet-grab` is a CSS class applied by default on the Leaflet map. Now if you inspect the DOM tree with Chrome/Firefox developer tools you see that a `.leaflet-dragging` class is added to the Map element when dragging. So I hope this will work.

martinRenou: As per in-person discussion with Sylvain and looking at the code, we think it will not work with the selector `.leaflet-dragging` unfortunately... Because the selector is used once, but it won't know if the `.leaflet-dragging` appears at some point. 

So if you want to implement this logic it must be done in JavaScript. On the `MapView` implementation you can listen for mouse events `mousemove`, and you could change the `style.cursor` value dynamically this way.

fitoprincipe: Ok, I see.. I tried, but didn't work.. as you expected.. do you think that should be changed now? I'll try to do my best, but I cannot guarantee..

martinRenou: Other idea: 

On the Python `Map` class you could have two styles:
```python
default_style = InstanceDict(MapStyle).tag(sync=True, **widget_serialization)
dragging_style = InstanceDict(MapStyle, default_value={'cursor': 'move'}).tag(sync=True, **widget_serialization)
``` 

Then on the JavaScript `LeafletMapView` class you could, on mouse events, choose which style is applied.
Something like:
```javascript
leaflet_events: function () {
    var that = this;
    this.obj.on('movestart', function (e) {
        this.model.get('default_style').unstyle();
        this.model.get('dragging_style').style();
    }
    this.obj.on('moveend', function (e) {
        this.model.get('dragging_style').unstyle();
        this.model.get('default_style').style();
    }
    // ...
}
```
I'm using the `style` and `unstyle` method from the style widget here, see https://github.com/jupyter-widgets/ipywidgets/blob/master/packages/base/src/widget_style.ts#L90.

fitoprincipe: It says: `that.model.get(...).unstyle is not a function` :thinking:

martinRenou: Oh sorry.. What `that.model.get('default_style')` returns is the `StyleModel`, not a `StyleView`...
Sorry for the time wasted, we are getting there hopefully...

There is a way to get the view from the model. Something like:
```javascript
that.model.get('default_style').views[0].then(function(style_view) {
    style_view.unstyle();
});
```
I start to really feel uncomfortable with those changes though...

fitoprincipe: Oh men, this was a head breaker for me! but I think I got it... 

I forgot about `dragging_style` and only used `style`...
I realized that `view` attribute wasn't an Array but an Object with a unique key...

``` javascript
this.obj.on('moveend', function (e) {
    var style_view = Object.keys(that.model.get('style').views)[0];
    that.model.get('style').views[style_view].then(function(view) {
        view.style();
    });
});
this.obj.on('movestart', function (e) {
    var style_view = Object.keys(that.model.get('style').views)[0];
    that.model.get('style').views[style_view].then(function(view) {
        view.unstyle();
    });
});
```
And this works fine! what do you think??

martinRenou: > I forgot about `dragging_style` and only used `style`

I do think we should have a `default_style` and a `dragging_style`. Because here you are using `style` only when dragging, which will not be obvious for other users.

> what do you think??

The fact that we are taking the first style view `Object.keys(that.model.get('style').views)[0]` is a bit weak, as nothing tells us that this first view is the one used by our `Map` (I think if you display the Map multiple times in your Notebook, your solution will work only for the first view of the Map).
Anyway, I'm happy to merge it as it is, and we can fix it later, before making a release.

The way to fix this would be to loop over all the views and call the `(un)style` function only for the view that respects the condition `view.parent === this`.

fitoprincipe: Hi @martinRenou, in my way of changing the code to use a `default_style` I found that when I change the attribute's name from `style` to anything else, it stops working, and inspecting I found that the object is created, but it has no views.. :thinking: maybe is for [this](https://github.com/jupyter-widgets/ipywidgets/blob/cd22809ed525d30ea20f37a65c49de5273de009c/packages/base/src/widget_style.ts#L75) line.. but I am just guessing..
Pantallazo-2019-04-26 08-08-59

martinRenou: Maybe you forgot to change the name in the deserializer? And in the `LeafletMapModel` property?

fitoprincipe: nope, I changed all

martinRenou: Do you mind pushing your changes?

fitoprincipe: Ok, I'll push it with the changes you proposed, with `dragging_style` and `default_style`

fitoprincipe:
This is the error message with the new push

Pantallazo-2019-04-26 09-29-23

js/src/Map.js
```diff
@@ -176,11 +197,31 @@ var LeafletMapView = utils.LeafletDOMWidgetView.extend({
...
-            }
+            };
+            var style_view = Object.keys(that.model.get('default_style').views)[0];
+            that.model.get('default_style').views[style_view].then(function(view) {
+                view.style();
+            });
+            var dragging_style_view = Object.keys(that.model.get('dragging_style').views)[0];
+            that.model.get('dragging_style').views[dragging_style_view].then(function(view) {
+                view.unstyle();
+            });
```
martinRenou: I guess you need to unstyle before styling, something like:
```javascript
var dragging_style_view = Object.keys(that.model.get('dragging_style').views)[0];
that.model.get('dragging_style').views[dragging_style_view].then(function(view) {
    // First remove the dragging_style
    view.unstyle();

    // Then apply the default style
    var style_view = Object.keys(that.model.get('default_style').views)[0];
    that.model.get('default_style').views[style_view].then(function(view) {
         view.style();
    });
});
```

And something equivalent on `movestart`: `unstyle` the `default_style` and `style` the `dragging_style`.

ipyleaflet/leaflet.py
```diff
@@ -2,9 +2,11 @@
 
 from ipywidgets import (
     Widget, DOMWidget, Box, Color, CallbackDispatcher, widget_serialization,
-    interactive
+    interactive, Style, CoreWidget
```
martinRenou: You don't use `CoreWidget` anymore you can remove it :)

js/src/Map.js
```diff
@@ -195,11 +197,31 @@ var LeafletMapView = utils.LeafletDOMWidgetView.extend({
...
             that.model.update_bounds().then(function() {
                 that.touch();
             });
         });
+
+        this.obj.on('movestart', function (e) {
+            var style_view = Object.keys(that.model.get('default_style').views)[0];
```
fitoprincipe: Still doesn't work.. same error as before. It cannot find the view :thinking:
For testing, just after this line I wrote:
``` javascript
console.log(style_view)
```
and returned `undefined`

martinRenou: Yeah I see that :( 

It looks like the `style` property must be named "style", nothing else.

fitoprincipe: Yes, I don't know why.. and also wondering if `dragging_style` will work (I guess not)

martinRenou: Maybe it is hardcoded somewhere in ipywidgets... I'll ask.

Every time we think we have a solution it does not work! Sorry for that.

martinRenou: Yes, it is actually some logic in the `DomWidgetView` class (`LeafletMapView inherits from it). See [this function](https://github.com/jupyter-widgets/ipywidgets/blob/1df5b73812a4186504c120e327389a8084a2f226/packages/base/src/widget.ts#L798). They create a view for the `style` property, only the `style` property, so if it has not the name `style` no view will be created.

Either we implement the same logic for `default_style` and `dragging_style`, or we come up with a solution not using the `Style` widget.

fitoprincipe: I though in a new solution, but (as always) I'm struggling.. 

This'd be in the python side:

``` python
style = InstanceDict(MapStyle).tag(sync=True, **widget_serialization)
dragging_style = Unicode('move').tag(sync=True, o=True)
default_style = Unicode('hand').tag(sync=True, o=True)
```
(I don't know what's the `o=True` for, but just in case...)

Then in `LeafletMapModel`:

``` javascript
...
style: null,
default_style: 'hand',
dragging_style: 'move',
...
```
I defined serializers for these two variables just in case (I'm not sure if it's needed)..

Then, in `leaflet_events`:
``` javascript
this.obj.on('movestart', function (e) {
    var style_model = that.model.get('style');
    var style_view = style_model.views[Object.keys(style_model.views)[0]]
    style_view.then(function(view){
        view.model.set('cursor', that.model.get('dragging_style'))
    })
});
this.obj.on('moveend', function (e) {
    ...
    var style_model = that.model.get('style');
    var style_view = style_model.views[Object.keys(style_model.views)[0]]
    style_view.then(function(view){
        view.model.set('cursor', that.model.get('default_style'))
    })
    ...
});
```
It works for the `movestart` event... but it doesn't for the `moveend`.. and I don't know why.. I tried to figure it out, but no luck :thinking:

martinRenou: Indeed, it would be a good approach. You don't need `o=True`, this is for specifying if a trait is a Leaflet option, `dragging_style` and `default_style` are not. You also don't need deserializers for `dragging_style` and `default_style`, only for `style`.

Do you have a JavaScript error? Or it just doesn't do anything?
Out of curiosity, do you mind trying the following:
```javascript
that.model.get('style').set('cursor', that.model.get('default_style'));
```
instead of
```javascript
    var style_model = that.model.get('style');
    var style_view = style_model.views[Object.keys(style_model.views)[0]]
    style_view.then(function(view){
        view.model.set('cursor', that.model.get('default_style'))
    })
```

fitoprincipe: I tried that before, but in the `moveend` event, and as it didn't work, I changed it.. but now I've tried to do it in the `movestart` event and works fine.. but it still does not work in the `moveend` event.. do I PR?

fitoprincipe: Also, what do you think if I change the variable names to be more specific to `cursor`?? because in the future another style could be modified.. like: `default_cursor` instead of `default_style` and `dragging_cursor` instead of `dragging_style`..

fitoprincipe: I don't get any JS errors.. it just doesn't behave as expected

martinRenou: > do I PR?

Yes please :)

> what do you think if I change the variable names to be more specific to cursor??

I agree. Or you could make `default_style` and `dragging_style` dictionaries containing the `cursor` key with its value.

fitoprincipe:
What is really confusing me is that if I change
``` javascript
this.obj.on('moveend', function (e) {
    ...
    that.model.get('style').set('cursor', that.model.get('default_style').cursor)
    ... 
    });
});
for this
``` javascript
this.obj.on('moveend', function (e) {
    ...
    console.log(that.model.get('default_style').cursor)
    that.model.get('style').set('cursor', 'hand'); 
    ... 
    });
});
it works.. and the log prints `hand`.. why doesn't it understand `hand` inside `set`??

**Comment by martinRenou** [comment]
Thas is indeed really weird :laughing:

**Comment by fitoprincipe** [comment]
We were so close.. :sob:

**Comment by martinRenou** [comment]
I'm so sorry that your first contribution to ipyleaflet is so difficult, believe me, it is not always like that!

I really don't get what is the difference between the two pieces of code you gave... I'll check this along the weekend

**Comment by fitoprincipe** [comment]
But if I print the new value of `cursor` after changing it using `console.log(that.model.get('style').get('cursor'))` it returns `hand`, which means it was correctly changed.. :thinking: It's like it's not updating the value somewhere else.. (like it'd be not listening..)

**Comment by fitoprincipe** [comment]
Ok @martinRenou, thank you for all your help! Have an excellent weekend =)

**Comment by martinRenou** [comment]
Yes, it is like the change event inside the `StyleModel` widget is not triggered, I tried triggering myself the event with `that.model.get('style').trigger('change:cursor')` without success. My next idea is to put `console.log` calls everywhere in the `StyleView` implementation.

Have a good week-end too :)

**Comment by fitoprincipe** [comment]
Hi @martinRenou, 
I've been doing some tests. As I haven't  set `ipywidgets` in development mode, to test `StyleView` I copied the rendered version of it (that I found in the webpack) into `Map.js` and pointed `LeafletMapStyleModel` to it. Then I did:

``` javascript
this.obj.on('movestart', function (e) {
    var views = Object.keys(that.model.get('style').views)
    that.model.get('style').views[views[0]].then(function(view){
        view.handleChange('cursor', that.model.get('dragging_style').cursor)
    })
});
this.obj.on('moveend', function (e) {
    var views = Object.keys(that.model.get('style').views)
    that.model.get('style').views[views[0]].then(function(view){
        view.handleChange('cursor', that.model.get('default_style').cursor)
    })
});
```
and surprise, it worked!.. but as soon as I point to the original `StyleView` that approach does not work :thinking:

martinRenou:
Did you have to go through the view, or did this work as well:
```javascript
that.model.get('style').set('cursor', that.model.get('default_style'));
```
(this approach is much cleaner/safer, as the approach with the view will fail if no view is created, and nothing tells us that the first view `view[0]` is ours).

Mmmmh, I don't get it :P You are saying that copying the code makes it work? As you did not modify the `StyleView` implementation it should be the same, right?

fitoprincipe:
Yes, the cleaner/safer approach works too.. and yes, I just copied the rendered `StyleView`...

``` javascript
var __extends = (this && this.__extends) || (function () {
    var extendStatics = Object.setPrototypeOf ||
        ({ __proto__: [] } instanceof Array && function (d, b) { d.__proto__ = b; }) ||
        function (d, b) { for (var p in b) if (b.hasOwnProperty(p)) d[p] = b[p]; };
    return function (d, b) {
        extendStatics(d, b);
        function __() { this.constructor = d; }
        d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
    };
})();

var StyleView = (function (_super) {
    __extends(StyleView, _super);
    function StyleView() {
        return _super !== null && _super.apply(this, arguments) || this;
    }
    StyleView.prototype.initialize = function (parameters) {
        this._traitNames = [];
        _super.prototype.initialize.call(this, parameters);
        // Register the traits that live on the Python side
        var ModelType = this.model.constructor;
        for (var _i = 0, _a = Object.keys(ModelType.styleProperties); _i < _a.length; _i++) {
            var key = _a[_i];
            this.registerTrait(key);
        }
        // Set the initial styles
        this.style();
    };
    StyleView.prototype.registerTrait = function (trait) {
        var _this = this;
        this._traitNames.push(trait);
        // Listen to changes, and set the value on change.
        this.listenTo(this.model, 'change:' + trait, function (model, value) {
            _this.handleChange(trait, value);
        });
    };
    StyleView.prototype.handleChange = function (trait, value) {
        // should be synchronous so that we can measure later.
        var parent = this.options.parent;
        if (parent) {
            console.log('changing', trait, 'to', value) // Testing log, to ensure it's using this code...
            var ModelType = this.model.constructor;
            var styleProperties = ModelType.styleProperties;
            var attribute = styleProperties[trait].attribute;
            var selector = styleProperties[trait].selector;
            var elements = selector ? parent.el.querySelectorAll(selector) : [parent.el];
            if (value === null) {
                for (var i = 0; i !== elements.length; ++i) {
                    elements[i].style.removeProperty(attribute);
                }
            }
            else {
                for (var i = 0; i !== elements.length; ++i) {
                    elements[i].style[attribute] = value;
                }
            }
        }
        else {
            console.warn('Style not applied because a parent view does not exist');
        }
    };
    StyleView.prototype.style = function () {
        for (var _i = 0, _a = this._traitNames; _i < _a.length; _i++) {
            var trait = _a[_i];
            this.handleChange(trait, this.model.get(trait));
        }
    };
    StyleView.prototype.unstyle = function () {
        var parent = this.options.parent;
        var ModelType = this.model.constructor;
        var styleProperties = ModelType.styleProperties;
        this._traitNames.forEach(function (trait) {
            if (parent) {
                var attribute = styleProperties[trait].attribute;
                var selector = styleProperties[trait].selector;
                var elements = selector ? parent.el.querySelectorAll(selector) : [parent.el];
                for (var i = 0; i !== elements.length; ++i) {
                    elements[i].style.removeProperty(attribute);
                }
            }
            else {
                console.warn('Style not removed because a parent view does not exist');
            }
        }, this);
    };
    return StyleView;
}(widgets.WidgetView));
```
I put I testing log in `handleChange`

fitoprincipe:
Oh my god.. I think a simple error (was my fault) broke everything.. The `hand` cursor does not exist, it's `grab`... I just changed it upon the latest commit and now it's working.. although (I think) it does not explain why copying the imported code into local code makes it work despite the error.. :thinking:

martinRenou:
Happy you found it! Yeah it is weird. So does it work now? :D

fitoprincipe:
Almost.. there is one thing that is not working.. testing the code in a notebook, when I first change `default_style` the cursor does not change.. I think `default_style` is not linked to `style`.. my first approach was making an `observe` function.. but did not work..

fitoprincipe:
I made a mistake in the `observe` function.. it works now.. you think is a good approach??

``` python
@observe('default_style')
def _observe_default_style(self, change):
    value = change['new']
    self.style = value
```

martinRenou:
Nop, I don't think it is a good approach. You should only change it if you are not dragging the map.
I think a better approach would be to implement it JavaScript side, with something like:

```javascript
this.model.on('change:default_style', function() {
    if (!that.dragging) { // I think you would need to define a boolean variable dragging somewhere, and set it to True when you start moving
        that.model.get('style').set('cursor', that.model.get('default_style')); // Maybe this should go in a custom function `update_style` or something equivalent if you have it on multiple places
    }
});
```

fitoprincipe:
Ok, it's working fine now, but before I PR, there is already a `dragging` attribute for `Map` https://github.com/jupyter-widgets/ipyleaflet/blob/74488d4699a5663fc28aabf94ebf08d956a30598/js/src/Map.js#L27
but I inspected inside `movestart` and `moveend` events, and does not change.. and also defaults to `true` 
what do I do? do I make a new attribute? replace that attribute? use that attribute (updating it on events)?

martinRenou:
Oh yeah sorry, the current `dragging` is a boolean saying if the user is allowed to drag the map or not. As it it would be an internal thing for JavaScript maybe name is `_dragging`

fitoprincipe:
That was my first approach! name `_dragging` and no sync.. but didn't get it right at first.. I'll try again =)

martinRenou:
Just `this._dragging`, no need to go through the `model`, if that's what didn't work ;)

fitoprincipe:
Oh, I've found a new bug.. this never ends! :laughing: .. I think we should define the possible values, because if a not allowed value is set, the behavior breaks (like it was happening before with the `hand` value).. I was thinking in following the pattern of `crs`.. but as `default_style` and `dragging_style` are `dict` I think is not possible.. should be another approach..

martinRenou:
I think it is fine for now if the value is not good, it would be up to the user to give a meaningful value. We could define [validators](https://traitlets.readthedocs.io/en/stable/using_traitlets.html#validation) for `default_style` and `dragging_style`. Or we can do the same pattern as for `crs` concerning the `MapStyle.cursor` attribute, and then find a way to use `MapStyle` validators for validating the `default_style` and `dragging_style` dictionaries. You can do it now if you want, but I'm happy to merge it without validators.

Also, do you think we should rename the `dragging` attribute `allow_dragging`? For better readability? We can do that in a separate PR. That would be a breaking change.

fitoprincipe:
> We could define validators for default_style and dragging_style

Yes, I like that. I'll have the time to do it on Thursday, but I'd like to do it before merging. Can I use multiple arguments in the validator decorator function? like

``` python
@validate('default_style', 'dragging_style', 'style')
def _valid_style(self, value):
    # rest of the code
```

> Do you think we should rename the dragging attribute allow_dragging?

I also like to define meaningful attributes, and I think that is a good change. I hope it doesn't break too much.

martinRenou:
> Can I use multiple arguments in the validator decorator function?

I guess :) but you would not be able to define the same validator for the `default_style` `dragging_style` dictionaries and for `style` Style (`style` already has a simple Unicode validation for the `cursor` value).
I think I would prefer a solution like:
```Python
allowed_cursor = ['grab', ...]

class MapStyle(Style, Widget):
    """ Map Style Widget """
    _model_name = Unicode('LeafletMapStyleModel').tag(sync=True)
    _model_module = Unicode("jupyter-leaflet").tag(sync=True)

    cursor = Enum(values=allowed_cursor, default_value='grab').tag(sync=True)
```
And if you turn `default_style` and `dragging_style` into `MapStyle` instances, then you would have validation. But you would need to change again the JavaScript code a bit. Something like:
```javascript
this.set('style', new_style);
```
instead of
```javascript
this.get('style').set('cursor', new_style.cursor);
```
in your `update_style` method (Don't forget deserializers if you do it this way).

If we do it this way, adding new CSS attributes to the MapStyle would be fairly easy, only a Python change and it would work. Because right now it is hardcoded JavaScript side that you can only change the `cursor` attribute.

fitoprincipe:
Hi @martinRenou ,
when I define `dragging_style` as a `MapStyle` instance, can I set the attribute `cursor` to `move`? how? Otherwise it takes the default value of `style` which is `grab` and stays with that value when dragging...

I tried 

``` python
dragging_style = InstanceDict(MapStyle, default_value=MapStyle({'cursor': 'move'})).tag(sync=True, **widget_serialization)
```
and
``` python
_default_dragging = MapStyle()
_default_dragging.cursor = 'move'
dragging_style = InstanceDict(MapStyle, default_value=_default_dragging).tag(sync=True, **widget_serialization)
```
but neither worked.. :thinking:

martinRenou:
What about 
```python
dragging_style = InstanceDict(MapStyle, default_value={'cursor': 'move'}).tag(sync=True, **widget_serialization)
```
?

fitoprincipe:
I tried that too.. doesn't work

martinRenou:
There is a `@default` decorator that you could use maybe https://traitlets.readthedocs.io/en/stable/using_traitlets.html#default-values-and-checking-type-and-value?

fitoprincipe:
That was it! I think we are done :clap: :tada:

martinRenou:
Awesome! I really do like the code, and it works like a charm! I hope you don't feel discouraged to contribute to ipyleaflet. I think that was a difficult task, as nobody did something equivalent in ipyleaflet.

I'll merge it if you are happy with it :)

fitoprincipe:
Of course! It was an awesome learning process! I feel more confortable to contribute now. Thank you for guiding me! Happy to be part of this project =)

martinRenou:
Thanks a lot! :)

PR #343, opened 2019-05-02
Change name of OpenTopoMap from 'OpenTopMap' to 'OpenTopoMap'

Swanson-Hysell:
I think that this was a typo given that OpenTopoMap is called using ```basemaps.OpenTopoMap```


martinRenou:
Thanks! :)

PR #344, opened 2019-05-03
Use latest version of leaflet-draw plugin

Kirill888:
Version 1.0.4, is currently most recent, this seems to fix some problems people
were having with DrawControl


Kirill888:
I also made a fix for jupyter lab

ipyleaflet/leaflet.py
```diff
@@ -759,6 +759,9 @@ class MapStyle(Style, Widget):
     """ Map Style Widget """
     _model_name = Unicode('LeafletMapStyleModel').tag(sync=True)
     _model_module = Unicode("jupyter-leaflet").tag(sync=True)
+
+    _model_module_version = Unicode(EXTENSION_VERSION).tag(sync=True)
```
martinRenou: Awesome, thanks!

martinRenou:
It works like a charm, thanks :)

Issue #341, opened 2019-05-01
GeoJSON style attribute makes no difference to object styling


Kirill888:
Using sample notebook from documentation:

https://ipyleaflet.readthedocs.io/en/latest/api_reference/geo_json.html

Changing any of the style parameters doesn't change the style of the GeoJSON layer. The documentation itself sets color to green, but there is no green anywhere.

Kirill888:
https://github.com/jupyter-widgets/ipyleaflet/blob/b3150a7becefeee306e2907053a6b6a7ef02b1b4/js/src/layers/GeoJSON.js#L20-L24

On line 24 above, I think `L.geoJson()` expects style to be a dictionary not a function.

Kirill888:
Note that this only for constructor supplied `style=` parameter, setting `style` after map was shown does change the styling.

Kirill888:
Also I have noticed that hover seems to reset style to some kind of default (blue), ignoring user supplied style changes.

deeplook:
I can confirm that. I'm setting color like this, but I always get the same bluesh color:

```python
GeoJSON(data=data, style={'color': 'red'})
```

deeplook:
After updating to 0.10.2 by styling issue disappeared. Thanks!

Issue #345, opened 2019-05-03
Resizing via `m.layout.height = ..` doesn't resize actual map


Kirill888:
Enclosing container for leaflet map remains always at 400px

I believe this is due to this css rule (line 2):

https://github.com/jupyter-widgets/ipyleaflet/blob/fffb6e5197b856e6dc11d8d7f2e420b33bb09be7/js/src/jupyter-leaflet.css#L1-L5

This is on `master` branch.

martinRenou:
Is the behavior on `master` different from the last release?

Kirill888:
I think it was changing size correctly before I switched to master.
I think solution is to use `100%` instead of `400px`

martinRenou:
Thanks for looking at it and for reporting the issue!

There were some recent changes on the layout, so it is possible that `master` is broken. Also, we need to check that ipyleaflet still works fine using a side-pane view or jupyterlab-sidecar, those features might also be broken.

Kirill888:
But then you have to set default `height` via usual widget layout mechanisms, otherwise you get empty maps within HBox/VBox

martinRenou:
This should be fixed on current master

AndrewAnnex:
I am seeing this issue with 0.13.6, although I have a complicated jupyterlab/conda setup that maybe interfering. if other jupyter installs and conda environments should not matter then something is wrong

PR #349, opened 2019-05-04
Fix styling problems for GeoJSON

Kirill888:
### GeoJSON style issue

There were two problems:

1. `_.extend(undefined, {..})` is `undefined`, so style would not apply to geojson objects without style defined.
2. New style would revert to original after every hover, since initial style was captured at
   construction and was never updated.

This addresses issue #341.

### DIFF
--- a/js/src/layers/GeoJSON.js
+++ b/js/src/layers/GeoJSON.js
@@ -16,9 +16,9 @@ var LeafletGeoJSONModel = featuregroup.LeafletFeatureGroupModel.extend({
 var LeafletGeoJSONView = featuregroup.LeafletFeatureGroupView.extend({
     create_obj: function () {
         var that = this;
-        var model_style = this.model.get('style');
         style = function (feature) {
-            return _.extend(feature.properties.style, model_style);
+            var model_style = that.model.get('style');
+            return _.extend(feature.properties.style || {}, model_style);
         }
         this.obj = L.geoJson(this.model.get('data'), {
             style: style,


Kirill888:
Also address map resize problem #345.

js/src/layers/GeoJSON.js
```diff
@@ -16,9 +16,9 @@ var LeafletGeoJSONModel = featuregroup.LeafletFeatureGroupModel.extend({
 var LeafletGeoJSONView = featuregroup.LeafletFeatureGroupView.extend({
     create_obj: function () {
         var that = this;
-        var model_style = this.model.get('style');
         style = function (feature) {
-            return _.extend(feature.properties.style, model_style);
+            var model_style = that.model.get('style');
+            return _.extend(feature.properties.style || {}, model_style);
```
martinRenou: Awesome, thanks for that!

js/src/Map.js
```diff
@@ -302,6 +302,11 @@ var LeafletMapView = utils.LeafletDOMWidgetView.extend({
         LeafletMapView.__super__.processPhosphorMessage.apply(this, arguments);
         switch (msg.type) {
             case 'resize':
+                var container = this.map_container;
+                if (container.clientHeight !== container.parentElement.clientHeight) {
+                    container.style.height = container.parentElement.clientHeight + 'px';
+                }
+
```
martinRenou: I'm not sure it's the right fix. I guess it would be cleaner to fix it in full CSS, which would mean making the `map_container` fill the widget container?

martinRenou: Maybe something like this fix #352? I basically replicate the "old" behavior, the one that you see on the last release. I fear that your fix only works on JupyterLab, and that it does not take into account the `map.layout` attribute.

js/src/layers/GeoJSON.js
Kirill888: Sorry, about multiple changes in one PR, I'll rebase this one to just GeoJSON, and do separate one for global `L`, and will leave out resize changes all together.

martinRenou: No, don't bother wasting time for that. It was just a comment for next time :)

js/src/Map.js
Kirill888: CSS still scares me, so yeah, this is probably not a good solution (but it did work in the notebook as well as lab), I removed this from this PR.

martinRenou:
Thanks a lot for those PRs!

PR #351, opened 2019-05-05
Expose fullscreen control/state

Kirill888:
Adding a two way `fullscreen` parameter, setting it to `True` puts map in to
full screen mode, and setting it to `False` exits full screen mode. Current
state is observable via this parameter as well.

## Browser Support

Typically browser will not allow to enter full screen if the event has not
been triggered by the user, so this can only be used in response to the user
action like a button click.

## Safari on macOS

Entering fullscreen mode via custom ui click doesn't work, looks like
safari expects fullscreen request and UI interaction that triggers it to happen
in the same animation frame and not just in "close enough in time". Observing
fullscreen state and exiting it does work however. And it's an important use
case, for example if user completes some input in full screen mode it might be
desirable to exit fullscreen at that point.

## Sample Notebook for testing

```python
import ipyleaflet as L
from ipywidgets import widgets as w

m = L.Map(zoom=2, layout=w.Layout(
    width="200px",
    height="200px",
))

btn = w.Button(description='Enter')
btn.layout.width = '6em'
m.add_control(L.WidgetControl(widget=btn, position='topright'))
m.add_control(L.FullScreenControl())

def on_btn_clk(btn):
    m.fullscreen = not m.fullscreen
    txt = 'Exit' if m.fullscreen else 'Enter'
    btn.description = txt

def on_fullscreen_change(e):
    txt = 'Exit' if e['new'] else 'Enter'
    btn.description = txt

m.observe(on_fullscreen_change, ['fullscreen'])
btn.on_click(on_btn_clk)
m
```

### DIFF
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -815,6 +815,7 @@ class Map(DOMWidget, InteractMixin):
     # zoom_animation = Bool(?).tag(sync=True, o=True)
     zoom_animation_threshold = Int(4).tag(sync=True, o=True)
     # marker_zoom_animation = Bool(?).tag(sync=True, o=True)
+    fullscreen = Bool(False).tag(sync=True, o=True)
 
     options = List(trait=Unicode).tag(sync=True)
 
--- a/js/src/Map.js
+++ b/js/src/Map.js
@@ -241,6 +241,9 @@ var LeafletMapView = utils.LeafletDOMWidgetView.extend({
                 location: that.model.get('location')
             });
         });
+        this.obj.on('fullscreenchange', function() {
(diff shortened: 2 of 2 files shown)


Kirill888:
I sort of copied behaviour of the plugin, it extends `Map` class with toggleFullscreen and isFullscreen methods, so even if you don't have fullscreen controller UI element, but have included this plugin, you can enter/exit full screen.

Still, I can change it to be a property of Controller rather than Map.

martinRenou:
I see what you mean. I guess it is fine like that then :)

We need to document clearly that changing the fullscreen attribute from Python will only work when coming from a callback. We can do that later, before the release.

PR #353, opened 2019-05-06
Use more robust check before using noConflict() mode

### LINKED ISSUE #337: Errors when `@jupyterlab/geojson-extension` and `jupyter-leaflet` are both enabled
Not sure where to report this issue as it is an interaction between

```
@jupyterlab/geojson-extension v0.18.1
jupyter-leaflet v0.10.1
```
using:

`jupyter-lab --version` is `0.35.4`

Individually both extensions work so long as the other one is disabled (I'm disabling at user level), but if both are enabled none work. They break with different error messages to the javascript console.

I'm not trying to use both in one notebook, separate notebooks

- Test notebook for GeoJSON
```python
from IPython.display import GeoJSON
GeoJSON({
    "type": "Feature",
    "geometry": {
        "type": "Point",
        "coordinates": [-118.4563712, 34.0163116]
    }
})
```
- Test notebook for ipyleaflet
```python
from ipyleaflet import Map
Map(zoom=2)
```

Looks like some sort of vendor package clashing issue.

Links:
- https://github.com/jupyter-widgets/ipyleaflet
- https://github.com/jupyterlab/jupyter-renderers/tree/master/packages/geojson-extension
- https://github.com/jupyterlab/jupyter-renderers/issues/189

Kirill888:
Geojson extension uses the same Leaflet lib version in jupyterlab, so calling `noConflict` is not needed when global `L` hasn't really changed, fixes #337.

### DIFF
--- a/js/src/index.js
+++ b/js/src/index.js
@@ -13,11 +13,12 @@ require('leaflet/dist/images/marker-icon.png');
 require('leaflet/dist/images/marker-icon-2x.png');
 
 // Export everything from jupyter-leaflet and the npm package version number.
-hasL = (typeof(window.L) != 'undefined');
+var _oldL = window.L;
 module.exports = require('./jupyter-leaflet.js');
 module.exports['version'] = require('../package.json').version;
 
-if (hasL) {
+// if previous L existed and it got changed while loading this module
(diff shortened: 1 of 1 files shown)


martinRenou:
Thanks for the fix :)

PR #355, opened 2019-05-06
Add layers control to map every time a layer is added

### LINKED ISSUE #348: Layers control disappear when adding a layer
Basically, if you first add the `LayersControl` first and then you add a layer, the controls disappear. To reproduce:

``` python
from ipyleaflet import *
m = Map(center=(52, 10), zoom=8, basemap=basemaps.Hydda.Full)
m
```
``` python
# Adding the control
m.add_control(LayersControl())
```
``` python
strata_all = basemap_to_tiles(basemaps.Strava.All)
m.add_layer(strata_all)
```

I found this (bad) behavior a while a go, but I though it could have something to do with me trying to contribute (modifying the code), but today I have tested it on binder (to be sure it's not my environment) and the bug appears too. There are no errors in the console, so it's hard to debug.

fitoprincipe:
fixes #348

### DIFF
--- a/js/src/controls/LayersControl.js
+++ b/js/src/controls/LayersControl.js
@@ -60,6 +60,8 @@ var LeafletLayersControlView = LeafletControlView.extend({
             }, {});
             that.obj = L.control.layers(baselayers, overlays);
             return that;
+        }).then(function() {
+               that.obj.addTo(that.map_view.obj);
         });
     }
 });


martinRenou:
Thanks a lot @fitoprincipe !

fitoprincipe:
Happy to help @martinRenou. I hope I can keep doing it =)

PR #359, opened 2019-05-13
Add no_wrap option for TileLayer

kgiacobbi:
This add the [noWrap](https://leafletjs.com/reference-1.5.0.html#gridlayer-nowrap) option for the TileLayer.

Whether the layer is wrapped around the antimeridian. If true, the TileLayer will only be displayed once at low zoom levels.


martinRenou:
Thanks a lot! :)

Issue #356, opened 2019-05-06
GridBox ipywidget in WidgetControl is broken


kgiacobbi:
Hi, i would like to add a GridBox to a WidgetControl but the GridBox is completly broken inside the WidgetControl.

```Python
from ipywidgets import GridBox, Layout, FloatRangeSlider, FloatText, Dropdown
from ipyleaflet import WidgetControl, Map

m = Map(zoom=1)

v1 = FloatText(layout = Layout(
                                width='auto', grid_area='v1'))

v2 = FloatText(layout = Layout(
                                width='auto', grid_area='v2'))

drop = Dropdown(layout = Layout(
                                width='auto', grid_area='drop'))

slider = FloatRangeSlider(layout = Layout(
                                width='auto', grid_area='slider'))


grid = GridBox(children=[v1,v2,drop, slider],
        layout=Layout(justify_content='center',
            width='500px',
            grid_template_rows='auto auto auto',
            grid_template_columns='80px 300px 80px',
            grid_template_areas='''
            "v1 . v2"
            ". slider ."
            " . drop ."

            ''')
       )
display(grid)

m.add_control(WidgetControl(widget=grid))

display(m)
```

kgiacobbi:
My GridBox in Jupyter lab:
![widget](https://user-images.githubusercontent.com/49714549/57282982-6c978300-70ae-11e9-9dbe-f2938590a9c4.png)

My GridBox in WidgetControl:
![widget_map](https://user-images.githubusercontent.com/49714549/57283027-86d16100-70ae-11e9-85c2-876ddfddbf46.png)

martinRenou:
I quickly took a look at it but could not find what the issue is. I suppose it is a CSS issue, there must be a CSS attribute missing in the widgetcontrol container. 
I don't really have time to look at it, if you would like to take a shot at fixing it that would be great!

kgiacobbi:
Hi,
I tried to look at the problem, but i can't found how to fix it.
If someone can try...
Because, for my internship, the widget has to be display well on the map. And for now, it's not.
Thanks

martinRenou:
Looking at your issue, I see a big difference between the content of the two grids. It looks like the style is not added:
Text box contained by the grid **outside** of the WidgetControl:
```html
<div class="p-Widget jupyter-widgets widget-inline-hbox widget-text" 
     style="width: auto; grid-area: v1 / v1 / v1 / v1;">...</div>
```
[...]

PR #364, opened 2019-05-17
Add workaround for box containers in WidgetControl

martinRenou:
This is a workaround for #356. 

WidgetControl contains a widget, if this widget also contains children (e.g VBox, GridBox) it appears that we need to trigger the 'displayed' event on them.

@SylvainCorlay and @jasongrout do you know why we need a `trigger('displayed', this)` on the child container in the first place? We do it [here](https://github.com/jupyter-widgets/ipyleaflet/blob/master/js/src/controls/WidgetControl.js#L79), **it appears that Layout properties are not added if we don't trigger the 'displayed' event**.

What I don't understand is that the 'displayed' event is not triggered in the [BoxView implementation in ipywidgets](https://github.com/jupyter-widgets/ipywidgets/blob/446d60864b3ef9e4615410aedf1dd59735736f52/packages/controls/src/widget_box.ts#L171).

This is the result of [this example](https://github.com/jupyter-widgets/ipyleaflet/issues/356#issue-440713999) without this fix:
widgetcontrol_before

And here is the expected result (with this fix):
widgetcontrol_fixed

**Note:** We should have the same issue with the current Popup implementation

### DIFF
--- a/js/src/controls/WidgetControl.js
+++ b/js/src/controls/WidgetControl.js
@@ -77,7 +77,19 @@ var LeafletWidgetControlView = LeafletControlView.extend({
                 // Trigger the displayed event of the child view.
                 this.displayed.then(() => {
                     this.widget_view.trigger('displayed', this);
+
                     this.widget_view.displayed.then(() => {
+                        // Workaround for box containers
+                        if (widget_model.get('children')) {
+                            _.each(widget_model.get('children'), (child) => {
+                                _.each(child.views, (view_promise) => {
+                                    view_promise.then((view) => {
+                                        view.trigger('displayed', this);
+                                    });
+                                });
+                            });
+                        }
+
                         this.updateLayout();
                         this.obj.setContent(view.el);
                     });


SylvainCorlay:
Please hold on to merge this PR. I want to review this.

martinRenou:
I don't think we should merge this anyway. There must be a better fix. Because the [BoxView implementation](https://github.com/jupyter-widgets/ipywidgets/blob/446d60864b3ef9e4615410aedf1dd59735736f52/packages/controls/src/widget_box.ts#L171) doesn't use anything like that for it to work. Try this, you will see that it works just fine:

```python
from ipywidgets import GridBox, Layout, FloatRangeSlider, FloatText, Dropdown, VBox

v1 = FloatText(layout = Layout(
                                width='auto', grid_area='v1'))

v2 = FloatText(layout = Layout(
                                width='auto', grid_area='v2'))

drop = Dropdown(layout = Layout(
                                width='auto', grid_area='drop'))

slider = FloatRangeSlider(layout = Layout(
                                width='auto', grid_area='slider'))

grid = GridBox(children=[v1,v2,drop, slider],
        layout=Layout(justify_content='center',
            width='500px',
            grid_template_rows='auto auto auto',
            grid_template_columns='80px 300px 80px',
            grid_template_areas='''
            "v1 . v2"
            ". slider ."
            " . drop ."

            ''')
       )
display(grid)

VBox((grid, ))
```

SylvainCorlay:
Fixed in #386.

PR #366, opened 2019-05-21
Adding an example of selecting GeoJSON features by clicking on the leaflet layer

mpdwulit:
Hi,

I have added a very simple example how to use leaflet to select Features from GeoJSON...

### DIFF
--- a/examples/Select-GeoJson.ipynb
+++ b/examples/Select-GeoJson.ipynb
@@ -1,0 +1,163 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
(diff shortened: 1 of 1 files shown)


martinRenou:
That is super nice, thanks a lot! Would you mind adding `pygeohash` as a binder dependency in the `environment.yml` file?

mpdwulit:
Ahh, yes I missed it. Thank you. 
The code has been refactored and the `pygeohash` dependency has been removed.

martinRenou:
Thanks :)

PR #385, opened 2019-06-19
Add selected layers in LayersControl

kgiacobbi:
Hi,
I added this attribute with the name of the selected layers in the LayersControl in JS.

But, in Python side, i cannot sync an attribute with it.

I tried this : 
`selected = List(trait=Unicode).tag(sync=True)` but it's not working.

Can you help me to add this feature ?

### DIFF
--- a/js/src/controls/LayersControl.js
+++ b/js/src/controls/LayersControl.js
@@ -61,6 +61,7 @@ var LeafletLayersControlView = LeafletControlView.extend({
             var pos = sel.indexOf(event.name);
             if(pos != -1){
                 sel.splice(pos, 1);
+                that.model.set('selected_overlays', sel);
             }
         });
 
--- a/js/src/controls/LayersControl.js
+++ b/js/src/controls/LayersControl.js
@@ -8,7 +8,9 @@ var LeafletControlModel = control.LeafletControlModel;
 var LeafletLayersControlModel = LeafletControlModel.extend({
   defaults: _.extend({}, LeafletControlModel.prototype.defaults, {
         _view_name: 'LeafletLayersControlView',
-        _model_name: 'LeafletLayersControlModel'
+        _model_name: 'LeafletLayersControlModel',
+        
+        selected: []
     })
 });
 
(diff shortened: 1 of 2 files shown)


js/src/controls/LayersControl.js
```diff
@@ -51,13 +73,22 @@ var LeafletLayersControlView = LeafletControlView.extend({
                 }
                 return ov;
             }, {});
+
+            var sel = that.model.get('selected');
             var overlays = views.reduce(function (ov, view) {
                 if (!(view.model.get("base")))
                 {
-                    ov[view.model.get("name")] = view.obj;
+                    var name = view.model.get("name");
```
martinRenou: Don't you need the same logic for baselayers as well?

kgiacobbi: If a base layer as the same name as an overlay layer, it will be a problem.
Or we do two distinct attributes.

martinRenou: Good idea, two distinct attributes makes a lot of sense. Something like `selected_layer` and `selected_overlays` ?

kgiacobbi: Something like that, but i think we will have a problem.
When you create a `LayersControl`, all the Layer are displayed. So they are all in `selected`. 
You check / uncheck some checkboxes and u decide to add a new `Layer` on the map. I'm afraid that `selected` will be completly wrong.

I saw something strange too. The object `layer` in JS have not an attribute `name` with the name of the layer in the event.
I don't know why because they have the attribute in the class definition.

martinRenou: > The object layer in JS have not an attribute name 

Ooh.. the layers in `LeafletJS` don't have a name. You only specify a name when you add a new layer to the control, isn't it?

js/src/controls/LayersControl.js
```diff
@@ -51,13 +77,22 @@ var LeafletLayersControlView = LeafletControlView.extend({
                 }
                 return ov;
             }, {});
```
martinRenou: You are also not initializing the `selected_base`

martinRenou:
Replaced by #435

PR #391, opened 2019-06-26
Fixes #367 : Options are now given to the JS control

### LINKED ISSUE #367: Changing position of some controls doesn't work
Hello, 
I saw that is not possible to change the position of those controls on a map. 
The FullScreenControl is always on the top left corner and LayersControl on the top right corner.
```
import ipyleaflet as lf
m = lf.Map()
m.add_control(lf.FullScreenControl(position='bottomleft'))
m.add_control(lf.LayersControl(position='bottomleft'))
m
```
capture_controls

I saw that the control has it option `position = 'bottomleft'` but i think the option is not passed at the JS side so it keeps the default value.

kgiacobbi:
Options were not given to JS objects

### DIFF
--- a/js/src/controls/FullScreenControl.js
+++ b/js/src/controls/FullScreenControl.js
@@ -19,7 +19,7 @@ var LeafletFullScreenControlView = LeafletControlView.extend({
     },
 
     create_obj: function () {
-        this.obj = L.control.fullscreen();
+        this.obj = L.control.fullscreen(this.get_options());
     },
 });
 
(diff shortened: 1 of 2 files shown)


martinRenou:
Good catch!! Thanks a lot! Isn't it why it was not respecting the `position`? See #367

kgiacobbi:
Yes, I saw that the MeasureControl could be moved and I noticed this difference.

martinRenou:
Awesome, thanks for fixing it :)

PR #393, opened 2019-06-28
Add ZoomControl has an Ipyleaflet.Control

kgiacobbi:
The ZoomControl is the zoom control added when you create a map with the option `zoom_control=True`.

The problem was you could not move it on the map.

Now you can do it :
```
import ipyleaflet as lf
m = lf.Map(zoom=1, zoom_control=False) #Deleting the original from the map
control = lf.ZoomControl(position='topright')
m.add_control(control)
m
```

### DIFF
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -855,7 +855,7 @@ class Map(DOMWidget, InteractMixin):
     default_style = InstanceDict(MapStyle).tag(sync=True, **widget_serialization)
     dragging_style = InstanceDict(MapStyle).tag(sync=True, **widget_serialization)
 
-    zoom_control = Bool(True).tag(sync=True)
+    zoom_control = Bool(False)
 
     @default('dragging_style')
     def _default_dragging_style(self):
@@ -892,6 +892,9 @@ class Map(DOMWidget, InteractMixin):
         super(Map, self).__init__(**kwargs)
         self.on_displayed(self._fire_children_displayed)
         self.on_msg(self._handle_leaflet_event)
+
+        if zoom_control:
+            self.add_control(ZoomControl())
 
     def _fire_children_displayed(self, widget, **kwargs):
         for layer in self.layers:
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -842,7 +842,6 @@ class Map(DOMWidget, InteractMixin):
     inertia_deceleration = Int(3000).tag(sync=True, o=True)
     inertia_max_speed = Int(1500).tag(sync=True, o=True)
     # inertia_threshold = Int(?, o=True).tag(sync=True)
-    zoom_control = Bool(True).tag(sync=True, o=True)
     attribution_control = Bool(True).tag(sync=True, o=True)
     # fade_animation = Bool(?).tag(sync=True, o=True)
     # zoom_animation = Bool(?).tag(sync=True, o=True)
@@ -855,6 +854,8 @@ class Map(DOMWidget, InteractMixin):
     style = InstanceDict(MapStyle).tag(sync=True, **widget_serialization)
     default_style = InstanceDict(MapStyle).tag(sync=True, **widget_serialization)
     dragging_style = InstanceDict(MapStyle).tag(sync=True, **widget_serialization)
+
+    zoom_control = Bool(True).tag(sync=True)
 
     @default('dragging_style')
     def _default_dragging_style(self):
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -893,7 +893,7 @@ class Map(DOMWidget, InteractMixin):
         self.on_displayed(self._fire_children_displayed)
         self.on_msg(self._handle_leaflet_event)
 
-        if zoom_control:
+        if self.zoom_control:
             self.add_control(ZoomControl())
 
     def _fire_children_displayed(self, widget, **kwargs):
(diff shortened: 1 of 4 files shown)


kgiacobbi:
We just need to change the way that it's added to the map

martinRenou:
> We just need to change the way that it's added to the map with zoom_control option

:+1: I think the way to go would be to not make it an option (_i.e._ remove the `o=True`). Pass `zoom_control = false` manually to the Leaflet Map. And create the `ZoomControl` instance in the `Map.__init__` manually, if `zoom_control is True`

kgiacobbi:
i will do that

deeplook:
That would be great to have because then I can put the zoom control below the search field in this example (otherwise it looks too weird): https://mybinder.org/v2/gh/deeplook/voila/master?urlpath=voila/render/notebooks/ipymaps.ipynb. :)

js/src/Map.js
```diff
@@ -56,7 +56,7 @@ var LeafletMapModel = widgets.DOMWidgetModel.extend({
         inertia_deceleration : 3000,
         inertia_max_speed : 1500,
         // inertia_threshold : int(?)
-        zoom_control : true,
+        zoom_control : false,
```
martinRenou: Actually, this line should be completely removed, no need to synchronize the `zoom_control` value with the front-end, only the Python side needs to know about this attribute.

js/src/Map.js
```diff
@@ -56,7 +56,7 @@ var LeafletMapModel = widgets.DOMWidgetModel.extend({
         inertia_deceleration : 3000,
```
martinRenou: When you actually create the Leaflet map (line 201 of this file), you should pass `zoom_control: false` to the Map constructor, so that Leaflet does not create a zoom control for us.

ipyleaflet/leaflet.py
```diff
@@ -846,6 +855,8 @@ class Map(DOMWidget, InteractMixin):
     default_style = InstanceDict(MapStyle).tag(sync=True, **widget_serialization)
     dragging_style = InstanceDict(MapStyle).tag(sync=True, **widget_serialization)
 
+    zoom_control = Bool(True).tag(sync=True)
```
martinRenou: No need to have `.tag(sync=True)`, this value will be used only Python side and won't be synchronized with the JavaScript side anymore.

ipyleaflet/leaflet.py
```diff
@@ -778,6 +778,16 @@ def clear_markers(self):
         self.send({'msg': 'clear_markers'})
```
martinRenou: You would need to change the `Map` constructor in this file. Maybe smth like:

```python
def __init__(self, *args, **kwargs):
    zoom_control = kwargs.get('zoom_control', True)
     if zoom_control:
        self.add_layer(ZoomControl())
    [...]
```

So that we keep the old behavior (creating the zoom control by default if `zoom_control is True`)

kgiacobbi:
```
 def __init__(self, **kwargs):
        super(Map, self).__init__(**kwargs)
        self.on_displayed(self._fire_children_displayed)
        self.on_msg(self._handle_leaflet_event)
        
        if self.zoom_control:
            self.add_control(ZoomControl())
```
kwargs is used in `super(Map, self).__init__(**kwargs)` so i can call self.zoom_control.

js/src/Map.js
```diff
@@ -198,7 +197,10 @@ var LeafletMapView = utils.LeafletDOMWidgetView.extend({
 
     create_obj: function () {
         return this.layoutPromise.then((views) => {
-            this.obj = L.map(this.map_container, _.extend({crs: L.CRS[this.model.get('crs')]}, this.get_options()));
+            options = this.get_options();
```
martinRenou: Maybe this is cleaner:

```javascript
var options = _.extend(
    {
        crs: L.CRS[this.model.get('crs')],
        zoom_control: false
    }, 
    this.get_options()
);

this.obj = L.map(this.map_container, options);
```

?

ipyleaflet/leaflet.py
```diff
@@ -846,6 +855,8 @@ class Map(DOMWidget, InteractMixin):
     default_style = InstanceDict(MapStyle).tag(sync=True, **widget_serialization)
     dragging_style = InstanceDict(MapStyle).tag(sync=True, **widget_serialization)
 
+    zoom_control = Bool(False)
```
martinRenou: We should add observers on `zoom_control` so that when the user sets the attribute later on with:

```python
m.zoom_control = False
```

we remove the control from the map.

We can do that later.

martinRenou: It should be `True` by default, if we want to keep the old behavior

kgiacobbi:
It's working:
zoom_control

ipyleaflet/leaflet.py
```diff
@@ -881,6 +892,15 @@ def __init__(self, **kwargs):
...
         self.on_msg(self._handle_leaflet_event)
+        
+    @observe('zoom_control')
+    def observe_zoom_control(self, change):
+        if change['new']:
+            self.add_control(ZoomControl())
+        else:
+            for control in self.controls:
+                if isinstance(control, ZoomControl):
+                    self.remove_control(control)
```
martinRenou: There might be a cleaner way to do it, by storing the ZoomControl instance as a Map attribute.

Anyway, I don't want to block your PR for that. We can clean it later :)

kgiacobbi: i can do the modification now.

PR #399, opened 2019-07-03
Load Stamen tiles over HTTPS

jtpio:
This will fix the following warnings when the notebook server is running behind HTTPS but the tiles are served over HTTP:

```
Loading mixed (insecure) display content “http://stamen-tiles-a.a.ssl.fastly.net/terrain/10/533/366.png” on a secure page
```


jtpio:
The other tile servers should also be using https if not already (some do but not all). It can be done in a follow-up PR.

martinRenou:
Thanks!

PR #400, opened 2019-07-03
Drop lab variables polyfill

SylvainCorlay:
This enables the dark theme in `voila`.


martinRenou:
Does it still work properly in classic notebook?

PR #403, opened 2019-07-04
GeoJSON sending feature and example

VasavanThiru:
Changing the GeoJSON.js file and send "feature" for on_hover method. 
You can also see how to use it in these notebooks examples :   GeoJson_EU_on_hover.ipynb  and Choropleth.ipynb

geojson_eu_on_hover


martinRenou:
Thanks !

PR #406, opened 2019-07-09
Geodata on hover

VasavanThiru:
Change the GeoData class in order to use on_hover.
You can also see how to use it in this notebooks example: GeoData_on_hover.ipynb

geodata_on_hover

### DIFF
--- a/examples/GeoData_on_hover.ipynb
+++ b/examples/GeoData_on_hover.ipynb
@@ -1,0 +1,109 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "scrolled": false
+   },
+   "outputs": [],
+   "source": [
+    "from ipyleaflet import Map, GeoData, basemaps, LayersControl\n",
+    "import geopandas\n",
+    "import json\n",
+    "\n",
+    "countries = geopandas.read_file(geopandas.datasets.get_path('naturalearth_lowres'))\n",
+    "rivers = geopandas.read_file(\"https://www.naturalearthdata.com/http//www.naturalearthdata.com/download/10m/physical/ne_10m_rivers_lake_centerlines.zip\")\n",
+    "\n",
+    "m = Map(center=(28.6019917,70.9121356), zoom = 3, basemap= basemaps.Esri.WorldTopoMap)\n",
+    "\n",
+    "geo_data = GeoData(geo_dataframe = countries,\n",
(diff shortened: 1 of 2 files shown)


ipyleaflet/leaflet.py
```diff
@@ -514,7 +514,7 @@ class GeoData(GeoJSON):
     geo_dataframe = Instance('geopandas.GeoDataFrame')
 
     def __init__(self, **kwargs):
-        super(GeoJSON, self).__init__(**kwargs)
+        super(GeoData, self).__init__(**kwargs)
```
martinRenou: Good catch!! Is it why the on_hover was not working?

VasavanThiru: yes definitely

VasavanThiru: Now it works ! It deserves a tweet !  ;)

examples/GeoData_on_hover.ipynb
```diff
@@ -0,0 +1,124 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
```
martinRenou: Do you mind removing the Notebook outputs?

VasavanThiru: Sure, it's clean now!

janbrus:
I installed ipywidgets using conda some weeks ago, and still got problem showing the hoover data running the example jypyter notebook GeoData_on_hover.ipynb. Is it possible that this error has come back?

davidbrochart:
I just tried, the notebook seems to work fine. Did you execute it entirely? The last cell enables the `on_hover`.

janbrus:
When I now replaced the leaflet.py i got from Conda-forge with the one here on Github it is working OK,

davidbrochart:
Then I suspect you have an installation issue. Could you try with a fresh conda environment?

janbrus:
I tried now and installed ipywidget in a new environment with the same result, but I am not an expert on this. To me it looks like leaflet.py July 29th size 70565 is the one one conda-forge that causes trouble. leaflet.py from October 19th size 78229 works ok.

martinRenou:
> When I now replaced the leaflet.py i got from Conda-forge with the one here on Github it is working OK,

Copying files around is the recipe to a broken installation, you should never do that :S 

What do you have when running `conda list | grep ipyleaflet` in the terminal?

janbrus:
ipyleaflet                0.13.3             pyh9f0ad1d_0    conda-forge

PR #407, opened 2019-07-12
Update examples/Numpy.ipynb 

jmp75:
* Updated USGS url for South America data download
* match changes in rasterio Dataset (`affine` property is obsoleted after rasterio 1.0)


martinRenou:
Thanks :)

Issue #414, opened 2019-08-07
DeprecationWarnings about traits


deeplook:
I keep getting warnings like these (using `traitlets 4.3.2`). Should these be addressed? 

```
.../lib/python3.7/site-packages/ipyleaflet/leaflet.py:587: DeprecationWarning: 
    Traits should be given as instances, not types (for example, `Int()`, not `Int`).
    Passing types is deprecated in traitlets 4.1.
    options = List(trait=Unicode).tag(sync=True)

.../lib/python3.7/site-packages/ipyleaflet/leaflet.py:434: DeprecationWarning:
    metadata {'trait': <traitlets.traitlets.Instance object at 0x121d827f0>} was set from the constructor.
    With traitlets 4.1, metadata should be set using the .tag() method, e.g., Int().tag(key1='value1', key2='value2')
    markers = Tuple(trait=Instance(Marker)).tag(sync=True, **widget_serialization)
```

martinRenou:
Thanks for the report. It should be easy to address.

martinRenou:
Fixed by #420, thanks @stephaniegott :)

xoolive:
Any plan to make a release which includes this fix?

martinRenou:
Doesn't 0.11.2 include this fix already?

xoolive:
doesn't seem so... 😞

martinRenou:
I'll have a look at it tomorrow 😊 I might make a release

xoolive:
The day after tomorrow is fine as well!

But it's weird: if I only import ipyleaflet, I don't get the warning. I only get if leaflet is imported through an extra library...

<img width="713" alt="Capture d’écran 2019-10-04 à 22 09 22" src="https://user-images.githubusercontent.com/1360812/66236917-001fbf00-e6f4-11e9-9692-44e5a1373ac1.png">

martinRenou:
`0.11.3` is out :)

You should be able to install it with pip already. The conda package will come today.

xoolive:
Thanks, there are indeed many less warnings!
But still some :)

In leaflet.py, there are few (3 or 4) such lines which raise a warning:
```python
Tuple(trait=Instance(Marker)).tag(sync=true, etc.)
```

to be replaced by 

```python
Tuple().tag(trait=Instance(Marker)sync=true, etc.)
```

[...]

PR #430, opened 2019-10-07
Fix some DeprecationWarning

xoolive:
Following discussion https://github.com/jupyter-widgets/ipyleaflet/issues/414#issuecomment-539007144

There is still one DeprecationWarning to fix in traittypes.Dataset (dtype should be set in a tag() method)

### DIFF
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -439,7 +439,7 @@ class LayerGroup(Layer):
     _view_name = Unicode('LeafletLayerGroupView').tag(sync=True)
     _model_name = Unicode('LeafletLayerGroupModel').tag(sync=True)
 
-    layers = Tuple(trait=Instance(Layer)).tag(sync=True, **widget_serialization)
+    layers = Tuple().tag(trait=Instance(Layer), sync=True, **widget_serialization)
 
     _layer_ids = List()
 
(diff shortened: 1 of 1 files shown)


martinRenou:
Concerning the Dataset trait, I wonder if you could not simply remove the `dtype` tag, it's set to its default value (`None`) anyway.

martinRenou:
Thanks a lot for the PR !!

xoolive:
I think it should be enough to remove it but in the traittypes package

martinRenou:
> I think it should be enough to remove it but in the traittypes package

Well, dtype seems to be used in the traittypes package, only we set it to `None` because I guess we don't want to force the user to use a specific dtype. I suppose removing it from ipyleaflet is just fine.

xoolive:
Well, I can't say: maybe it's fine, but it doesn't fix the warning :disappointed:

martinRenou:
I don't have those warnings... Which traitlets version are you using?

xoolive:
traitlets==4.3.2

But I don't get the warnings on a regular import, only when imported through an entry point:

```python
# something along
for entry_point in pkg_resources.iter_entry_points("personal.stuff"):
    entry_point.load()
```

martinRenou:
I can merge this one as it is right now as it fixes some more warnings.

Thanks a lot for this PR! :)
I'll investigate concerning this Dataset thing

xoolive:
Thanks, about the extra warning, see https://github.com/jupyter-widgets/traittypes/pull/41

PR #431, opened 2019-10-14
Add substitute_layer to LayerGroup class

joshwarren:
This maintains consistency with Map class.


martinRenou:
I wonder if `LayerGroup` and `Map` should not have the same base class in order to factorize this code.  Could be an idea for future improvements.

Thanks again for your contribution!

PR #443, opened 2019-11-20
Add spinner when tiles are loading

davidbrochart:
API is the following:
```python
from ipyleaflet import Map

m = Map(center=(46, 14), zoom=5, zoom_control=True)
m

l = m.layers[0]
l.show_loading = True
```

### DIFF
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -876,8 +876,6 @@ class Map(DOMWidget, InteractMixin):
     attribution_control = Bool(True)
     attribution_control_instance = AttributionControl(position='bottomright')
 
-    show_progress = Bool(False)
-
     @default('dragging_style')
     def _default_dragging_style(self):
         return {'cursor': 'move'}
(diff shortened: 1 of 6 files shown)


ipyleaflet/leaflet.py
```diff
@@ -875,6 +876,8 @@ class Map(DOMWidget, InteractMixin):
     attribution_control = Bool(True)
     attribution_control_instance = AttributionControl(position='bottomright')
 
+    show_progress = Bool(False)
```
martinRenou: Is this a leftover?

davidbrochart: Right, thanks!

martinRenou:
Just a few comments:
- The spinner is a bit big
- It would be nicer if the spinner respects the current JupyterLab theme
- It might also be nicer to position the spinner on the top-right corner of the Map instead of the middle of it. We can play with the `top` and `left` spinner options

We can address all of this in separate PRs :)

PR #449, opened 2019-12-15
Add basemaps with creds

deeplook:
This adds an example for using configurable basemaps with API credentials. Probably it could be structured better in a way to bundle modules for building tile URLs for separate providers inside some subdirectory, but I didn't want to go too far with this and rather solicit some early comments about the way to do this best...


deeplook:
I see that fb649e3c72f7d58f4dc82ef9643397542fbce9c0 should not have been in this, sorry!

martinRenou:
Would you mind removing it?

deeplook:
Abondoning this in favour of https://github.com/jupyter-widgets/ipyleaflet/pull/450.

PR #450, opened 2019-12-20
Add basemap with creds

deeplook:


SylvainCorlay:
Sorry @deeplook this needs a (very simple) rebase since we added the copyright headers.

deeplook:
@SylvainCorlay I've resolved the conflict. Ok now?

deeplook:
I think this should be it.

davidbrochart:
@deeplook this doesn't work with python3: `ModuleNotFoundError: No module named 'here'`.
In python3 it should be a relative import: `from .here import ...`

deeplook:
@davidbrochart Oups, maybe some copying artifact. Should be solved.

ipyleaflet/basemaps.py
```diff
@@ -245,5 +247,20 @@
             min_zoom = 1,
             max_zoom = 18
         )
+    ),
+    HERE = Bunch(
```
martinRenou: Maybe it would be a good idea to put those basemaps in a different bunch? I would say in `here.py`? So that there is a clear separation between basemaps without creds and basemaps with creds?

ipyleaflet/here.py
```diff
@@ -0,0 +1,109 @@
...
+    # Or use legacy APP_CODE and APP_ID if provided, instead.
+    app_code = kwargs.get("app_code", None) or os.getenv("HEREMAPS_APP_CODE", None)
+    app_id = kwargs.get("app_id", None) or os.getenv("HEREMAPS_APP_ID", None)
+    if app_code and app_id:
+        tiles_url = tiles_url.format(api_url="maps.api.here.com")
+        tiles_url += f"&app_id={app_id}&app_code={app_code}"
+        return tiles_url
+
+    raise ValueError(
+        "HERE APIKEY or APP_CODE/APP_ID needed to be in params or env. vars.")
```
martinRenou: Maybe add here a comment saying how to get credentials? I see it's already in the docstring, but errors are also good documentation :)

deeplook: Added a pointer to use help() to get more info from the docstring. Otherwise it's just too long for the error message.

ipyleaflet/basemaps.py
deeplook: Done. I still think a new subdir like `providers` or so would be a good idea, but I have a hard time to come up with an appropriate name. `providers` sounds too general while anything else I can think of would include `basemaps` and/or `credentials` or `proprietary` which sounds too long. Any idea?

ipyleaflet/here.py
```diff
@@ -84,11 +86,12 @@ def build_tiles_url(**kwargs):
         return tiles_url
 
     raise ValueError(
-        "HERE APIKEY or APP_CODE/APP_ID needed to be in params or env. vars.")
+        "HERE APIKEY or APP_CODE/APP_ID needed in params or env. variables. "
+        "For more information please type 'help(build_tiles_url'.")
```
martinRenou: There seem to be a missing closing parenthesis in the message: `help(build_tiles_url)`

ipyleaflet/here.py
```diff
@@ -99,11 +102,28 @@ def basemap_to_tiles(**kwargs):
...
         min_zoom = kwargs.get("min_zoom", 1),
         max_zoom = kwargs.get("max_zoom", 18),
         attribution = '&copy; <a href="https://here.com">HERE.com</a>',
         name = "HERE",
+        **kwargs
     )
-    return params  # better use TileLayer?
+
+
+basemaps = Bunch(
```
martinRenou: Comment: I wonder if we should not change this name to something like `heremaps`? And do something equivalent for future providers in this directory?

ipyleaflet/here.py
martinRenou: Agree :)

ipyleaflet/here.py
deeplook: But then it would be a different name to memorize for each provider.

ipyleaflet/here.py
deeplook: Well spotted. Fixed.

ipyleaflet/basemaps.py
martinRenou: `external_providers`? Maybe something general is not too bad. I would like to add maki icons in this directory later so something containing `basemaps` in the name would not make sense in this case.

ipyleaflet/here.py
martinRenou: Sure, I agree

ipyleaflet/basemaps.py
deeplook: So `external` it is now...

deeplook: Done.

ipyleaflet/external/here.py
```diff
@@ -0,0 +1,129 @@
...
+        maptype="aerial", tiletype="maptile", scheme="satellite.day"),
+    TrafficDay = basemap_to_tiles(
+        maptype='traffic', tiletype='traffictile', scheme="normal.day"),
+    TrafficNight = basemap_to_tiles(
+        maptype='traffic', tiletype='traffictile', scheme="normal.night"),
+    TransitDay = basemap_to_tiles(
+        maptype='base', tiletype='maptile', scheme="normal.day.transit"),
+    TransitNight = basemap_to_tiles(
+        maptype='base', tiletype='maptile', scheme="normal.night.transit"),
+    Parameterized = basemap_to_tiles,
```
martinRenou: What is this for? @deeplook

martinRenou:
Thanks a lot!

PR #459, opened 2020-01-09
Add leaflet-ant-path plugin

davidbrochart:

### DIFF
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -399,7 +399,7 @@ class AntPath(VectorLayer):
     locations = List().tag(sync=True)
 
     # Options
-    use = Unicode('polyline').tag(sync=True, o=True)
+    use = Enum(values=['polyline', 'polygon', 'rectangle', 'circle'], default_value='polyline').tag(sync=True, o=True)
     delay = Int(400).tag(sync=True, o=True)
     weight = Int(5).tag(sync=True, o=True)
     dash_array = List([10, 20]).tag(sync=True, o=True)
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -392,6 +392,25 @@ class Path(VectorLayer):
     opacity = Float(1.0, min=0.0, max=1.0).tag(sync=True, o=True)
 
 
+class AntPath(VectorLayer):
+    _view_name = Unicode('LeafletAntPathView').tag(sync=True)
+    _model_name = Unicode('LeafletAntPathModel').tag(sync=True)
+
+    locations = List().tag(sync=True)
+
+    # Options
+    use = Unicode('polyline').tag(sync=True, o=True)
+    delay = Int(400).tag(sync=True, o=True)
+    weight = Int(5).tag(sync=True, o=True)
+    dash_array = List([10, 20]).tag(sync=True, o=True)
+    color = Color('#0000FF').tag(sync=True, o=True)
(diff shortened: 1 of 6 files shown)


ipyleaflet/leaflet.py
```diff
@@ -392,6 +392,25 @@ class Path(VectorLayer):
...
 
 
+class AntPath(VectorLayer):
+    _view_name = Unicode('LeafletAntPathView').tag(sync=True)
+    _model_name = Unicode('LeafletAntPathModel').tag(sync=True)
+
+    locations = List().tag(sync=True)
+
+    # Options
+    use = Unicode('polyline').tag(sync=True, o=True)
```
martinRenou: Maybe use an `Enum` so you can validate the value. `Enum(values=['polyline', 'polygon', 'circle'], default_value='polyline')`

js/src/layers/AntPath.js
```diff
@@ -0,0 +1,52 @@
...
+var L = require('../leaflet.js');
+var vectorlayer = require('./VectorLayer.js');
+var LeafletVectorLayerView = vectorlayer.LeafletVectorLayerView;
+var LeafletVectorLayerModel = vectorlayer.LeafletVectorLayerModel;
+var {antPath} = require('leaflet-ant-path');
+
+var LeafletAntPathModel = LeafletVectorLayerModel.extend({
+    defaults: _.extend({}, LeafletVectorLayerModel.prototype.defaults, {
+        _view_name : 'LeafletAntPathView',
+        _model_name : 'LeafletAntPathModel'
```
martinRenou: You are missing the model default values (use: 'polyline', radius: 10, etc)

js/src/layers/AntPath.js
```diff
@@ -0,0 +1,52 @@
...
+    })
+});
+
+var LeafletAntPathView = LeafletVectorLayerView.extend({
+    create_obj: function () {
+        this.obj = antPath(this.model.get('locations'), this.get_ant_options());
+    },
+
+    model_events: function () {
+        LeafletAntPathView.__super__.model_events.apply(this, arguments);
```
martinRenou: You don't seem to have added support for changing the `locations` dynamically.

js/src/layers/AntPath.js
```diff
@@ -0,0 +1,52 @@
...
+        var key;
+        var o = this.model.get('options');
+        for (var i=0; i<o.length; i++) {
+            key = o[i];
+            this.listenTo(this.model, 'change:' + key, function () {
+                this.obj.setStyle(this.get_ant_options());
+            }, this);
+        }
```
martinRenou: Could this for-loop could be replaced by:

```javascript
this.model.on_some_change(this.model.get('options'), () => {
    this.obj.setStyle(this.get_ant_options());
});
````
?

ipyleaflet/leaflet.py
davidbrochart: Good point, thanks.

js/src/layers/AntPath.js
davidbrochart: Right!

js/src/layers/AntPath.js
davidbrochart: Indeed.

js/src/layers/AntPath.js
davidbrochart: Yes, that works, thanks!

davidbrochart:
I think so!

Issue #455, opened 2020-01-07
"loading" as an attribute of TileLayer?


fitoprincipe:
Hi all,

the new `show_loading` attribute added recently to the `TileLayer` class shows an animation each time any tile is loading, it is fine, but wouldn't it be better if a `loading` property (read only attribute) is set to `True` each time any tile is loading? Therefore we could make use of it our own way.

martinRenou:
Hi! There is an `on_load` event that is triggered when the tiles finished loading. Something like:

```Python
def on_load(**kwargs):
    # Do something
    pass

my_tile_layer.on_load(on_load)
```

Maybe that could suit your needs?

fitoprincipe:
Hi @martinRenou, I tried to make a separate widget that shows a red bar when loading tiles and a green bar when all tiles are loaded, but when you move new tiles load but I "catch" this so I can make the bar red again. I though of using the map's events, but then I though it would be better to just have a `loading` attribute.

martinRenou:
I totally agree. It should not be too difficult to add. Would you like to take a shot? :)

fitoprincipe:
Yes, sure =) I will try

fitoprincipe:
Ok, I need some help @martinRenou ..

in `leaflet.py`:

``` python
class TileLayer(RasterLayer):
    ...
    loading = Bool(False).tag(sync=True, o=True)
```
then in `TileLayer.js`:

``` javascript
var LeafletTileLayerView = LeafletRasterLayerView.extend({
    leaflet_events: function () {
        LeafletTileLayerView.__super__.leaflet_events.apply(this, arguments);
        var that = this;
        this.obj.on('loading', function (e) {
            that.model.set('loading', true);  // this is how the attr center of Map is updated
            if (that.model.get('show_loading'))
                that.spinner = new Spinner().spin(that.map_view.el);
        });
        this.obj.on('load', function() {
           that.model.set('loading', false); // set it to false when finished loading..
            that.send({
                event: 'load'
            });
            if (that.model.get('show_loading'))
                that.spinner.stop();
        });
    },
})
```
Then I tested it with a listener:

``` python
m = ipyleaflet.Map()
lay = TileLayer(url='https://{s}.tile.opentopomap.org/{z}/{x}/{y}.png')
m.add_layer(lay)
def ob(change):
    print('loading')
lay.observe(ob, names='loading')
```
[...]

PR #460, opened 2020-01-09
add loading attribute to TileLayer

fitoprincipe:
There is code of another approach I made. We can clean it up later.


martinRenou:
This should fix #455

js/src/layers/TileLayer.js
```diff
@@ -23,7 +23,13 @@ var LeafletTileLayerModel = LeafletRasterLayerModel.extend({
         detect_retina : false,
         tms: false,
         show_loading: false,
-    })
+        loading: false,
+    }),
+    
+    update_loading: function(value) {
```
martinRenou: You can remove this :)

Issue #83, opened 2017-11-14
Support same GeoJSON conventions as GitHub


rsignell-usgs:
It would be great of ipyleaflet supported the same GeoJSON conventions as GitHub: https://help.github.com/articles/mapping-geojson-files-on-github/

This would allow Ipyleaflet users to easily change marker colors and styles, fixing problems like https://github.com/reproducible-notebooks/ERDDAP_timeseries_explorer/issues/3.

Simple example:  
https://github.com/rsignell-usgs/dc-wifi-social/blob/master/bars.geojson
produces map that looks like this:
![2017-11-14_8-45-05](https://user-images.githubusercontent.com/1872600/32782997-415ead7e-c918-11e7-9842-638b284ee862.jpg)

While the same GeoJSON in Ipyleaflet:
```python
import urllib.request
import json
url = 'https://raw.githubusercontent.com/rsignell-usgs/dc-wifi-social/master/bars.geojson'
req = urllib.request.Request(url)
r = urllib.request.urlopen(req).read()
data = json.loads(r.decode('utf-8'))
center = [38.9, -77.05]
zoom = 12
map = ipyl.Map(center=center, zoom=zoom, layout=ipyl.Layout(width='650px', height='350px'))
feature_layer = ipyl.GeoJSON(data=data)
map.layers = [map.layers[0], feature_layer]
map
```
produces a map that looks like this:
![2017-11-14_8-45-40](https://user-images.githubusercontent.com/1872600/32783087-8d9f87d0-c918-11e7-9e4a-8638b073337e.jpg)

ocefpaf:
I believe that that is specific to `mapbox`, not sure if it is possible to do it with `leaflet`.

One way to do it with `leaflet` is with https://github.com/albburtsev/Leaflet.geojsonCSS

jasongrout:
You can create custom marker icons in leaflet: http://leafletjs.com/examples/custom-icons/

That's not hooked directly up to the geojson data, but there may be something possible there.

ocefpaf:
> That's not hooked directly up to the geojson data, but there may be something possible there.

`Leaflet.geojsonCSS` is a plugin to read the custom-icons info from the GeoJson data, but one would need to convert from the mapbox style to leaflet's style.

jasongrout:
For completeness, I think here are the docs about those icons in mapbox: https://www.mapbox.com/help/markers-js/#style-markers-with-simplestyle

tanay27:
[...]

PR #462, opened 2020-01-09
Add support for font-awesome icons in Markers

martinRenou:
icons2

Related to https://github.com/jupyter-widgets/ipyleaflet/issues/83, it does not add automatic support in the GeoJSON layer though.

### DIFF
--- a/examples/AwesomeIcons.ipynb
+++ b/examples/AwesomeIcons.ipynb
@@ -1,0 +1,95 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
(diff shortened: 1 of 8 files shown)


martinRenou:
@davidbrochart That would be awesome if you could look at this one too! :)

I think we should do a release tomorrow with all those new changes/bug fixes

davidbrochart:
Awesome!

martinRenou:
I guess I can take this as approval :P

davidbrochart:
:+1:

PR #465, opened 2020-01-10
TileLayer loading attribute example and doc

fitoprincipe:


martinRenou:
Thanks a lot! Is it ready to be merged?

fitoprincipe:
If you agree with the name of the example notebook and with the comment on docs, yes! =)

fitoprincipe:
I don't understand why it takes the first two commits that were already merged :thinking:

docs/source/api_reference/tile_layer.rst
```diff
@@ -61,4 +61,5 @@ opacity            1.0
 visible            True
 no_wrap            False
 show_loading       False
+loading            False (dynamically updated)
```
martinRenou: This makes me think that the `loading` attribute should be specified as read-only. Something like:

```python
loading = Bool(False, read_only=True).tag(sync=True)
```

So that users would have an error when they try to assign to it. Could be in a separate PR though, as you wish.

martinRenou:
> I don't understand why it takes the first two commits that were already merged

That's fine, I can squash commits when merging :)

I would suggest creating branches next time instead of committing to the master branch. That would prevent this :)

docs/source/api_reference/tile_layer.rst
fitoprincipe: Yes, I though of that. I'll do it right now and see if doesn't break anything. If it doesn't, then I'll commit again.

fitoprincipe: `Cannot find module "leaflet.awesome-markers"` in JS console... :thinking:

martinRenou: Oh yes, it's a new dependency that has been added. You need to run `npm install` to install it.

fitoprincipe:
> I would suggest creating branches next time instead of committing to the master branch. That would prevent this :)

OK!

npm-debug.log
```diff
@@ -0,0 +1,95 @@
+0 info it worked if it ends with ok
```
martinRenou: You might have committed this file by mistake?

fitoprincipe: that is weird, did not appear when I did `git diff`, besides, I did not create it manually, so it is created automatically.. but it is not ignored in `.gitignore`..

fitoprincipe: can you help me on this? I did:

``` bash
git rm --cached npm-debug.log
git add -A
git commit -m 'remove file commited by mistake'
git push
```
But did not worked as expected :thinking:

martinRenou: Which part did not work?

fitoprincipe: I don't see my commit here :thinking:... I pushed it

martinRenou: Are you sure `git push` did not give an error?

fitoprincipe:
how does it look now? By the way, why the npm log file was created on that folder?

martinRenou:
I guess because you ran 'npm run build' from the root directory at some point..

Thanks for the PR!

fitoprincipe:
May be I run `npm run watch` accidentally in that folder. OK. Thanks! =)

PR #476, opened 2020-01-16
Add zoom_control_instance default value

martinRenou:
This prevents an error with `self.zoom_control_instance` not existing.

Same fix for the attribution control.


SylvainCorlay:
LGTM

PR #480, opened 2020-01-18
Fix HERE basemaps configuration

deeplook:


ipyleaflet/external/here.py
```diff
@@ -25,6 +25,9 @@
 from ipyleaflet import TileLayer
 
 
+attribution = '&copy; <a href="https://here.com">HERE.com</a>'
```
martinRenou: What is the reason for using a global variable now?

ipyleaflet/external/here.py
```diff
@@ -90,24 +93,30 @@ def build_tiles_url(**kwargs):
...
-     'attribution': '&copy; <a href="https://here.com">HERE.com</a>',
-     'name': 'HERE'}
+    >>> basemap_to_tiles(basemaps.Default, config={"apikey": "foobar"}).url
+    'https://4.base.maps.ls.hereapi.com/maptile/2.1/maptile/newest/normal.day/{z}/{x}/{y}/256/png8?lg=eng&apiKey=foobar'
     """
+    basemap2 = {}
+    basemap2.update(basemap)
+    basemap2.update(config)
     return TileLayer(
-        url=build_tiles_url(**basemap, **kwargs),
```
martinRenou: I would have a personal preference for:
```python
url=build_tiles_url(**basemap, **config, **kwargs)
```
Instead of using a local variable.

Although feel free to ignore this comment if you like it the way you did it.

martinRenou:
You need to rebase your PR in order to remove all the previous commits:
```
git checkout master
git pull https://github.com/jupyter-widgets/ipyleaflet.git master

git checkout fix_here_basemaps
git rebase -i master
```

Should do the trick.

Then you need to force push your branch:
```
git push -f origin fix_here_basemaps
```

ipyleaflet/external/here.py
deeplook: Nothing special, a left-over from the original version. Will make it a fixed string.

ipyleaflet/external/here.py
deeplook: Ok. for the `**config`, but the `**kwargs` are meant for `TileLayer(...)` only or am I missing something?

ipyleaflet/external/here.py
deeplook: Done.

ipyleaflet/external/here.py
martinRenou: Yes :)

deeplook: Ok now?

deeplook: Not ready yet.

deeplook: Changed to `url=build_tiles_url(**{**basemap, **config})`.

PR #495, opened 2020-01-31
Patch LeafletJS for replacing tiles instead of removing/adding when changing the tiles URL

martinRenou:
Before:
before

After:
after


davidbrochart:
I looks much nicer.

martinRenou:
This needs some more testing. Some tiles are not removed and I don't know yet why.

martinRenou:
I used @barryhunter's patch instead. It looks more reliable than my first version (see https://github.com/Leaflet/Leaflet/issues/6659)

PR #502, opened 2020-02-07
Remove jlab pinning on binder

martinRenou:


martinRenou:
This will fix #328

jtpio:
Maybe also combine both into one command? To trigger only one build.

`jupyter labextension install @jupyter-widgets/jupyterlab-manager jupyter-leaflet`

jtpio:
Can be tested with (after the build is finished): https://mybinder.org/v2/gh/martinRenou/ipyleaflet/jlab_pinning?urlpath=/lab/tree/examples

PR #505, opened 2020-02-21
Sync Path's fill_color with color

yudai-nkt:
This PR fills circle marker and other objects with the same color as the stroke color except for the opacity.

```python
from ipyleaflet import CircleMarker, Map

m = Map()
m.add_layer(CircleMarker(color='#d9a89c', fill_opacity=0.5, radius=100))
m
```

|  w/o PR  |  w/ PR |
| ---- | ---- |
| before | after |

In Leaflet, the value of `Path`'s `fillColor` option defaults to its `color` value as is documented [here](https://leafletjs.com/reference-1.6.0.html#path-fillcolor). So I modified ipyleaflet accordingly.

IMHO this would be the behavior users find comfortable (at least I do), but feel free to close if ipyleaflet has some background not to be aligned with Leaflet.

### DIFF
--- a/docs/source/api_reference/circle.rst
+++ b/docs/source/api_reference/circle.rst
@@ -36,7 +36,7 @@ color             "#0033FF"          Stroke color
 opacity           1.0                Stroke opacity
 weight            5                  Stroke width in pixels
 fill              True               Whether to fill the circle or not
-fill_color        "#0033FF"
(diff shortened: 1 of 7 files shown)


martinRenou:
And thanks a lot for updating the docs accordingly!

PR #510, opened 2020-03-04
Deploy extension package for JLab

martinRenou:


martinRenou:
@mariobuikhuizen I am trying to achieve the same thing as you did here: https://github.com/mariobuikhuizen/ipyvue/blob/master/setup.py#L34. Basically installing the `tgz` so that JupyterLab sees it and asks me to rebuild from the interface. 

But it does not seem to work. The `tgz` file is not put under `share/jupyter/lab/extensions`. 

Would you mind looking at it? Maybe you'll directly see what I'm doing wrong.

martinRenou:
Apparently `pip install -e .` does not install the data files. It does work without the `-e`.

mariobuikhuizen:
Great!

martinRenou:
Thanks Mario for doing this in `ipyvue`, it's a great improvement and should be done in all widgets packages IMO

mariobuikhuizen:
> Thanks Mario for doing this in `ipyvue`, it's a great improvement and should be done in all widgets packages IMO

Thanks go to @maartenbreddels, I've copied it from him on his suggestion :-).  Mostly to not break existing users of ipyvuetify when ipyvue was extracted from it.

maartenbreddels:
And I learned it from @vidartf :)

PR #511, opened 2020-03-04
Add Travis installation check

### LINKED ISSUE #377: Enable CI
We should enable CI with Travis. Even if we have no test **yet**, it would prevent this kind of issues #375.

martinRenou:
This will fix #377

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,29 @@
+language: generic
+os:
+  - linux
+  - osx
+env:
+ matrix:
+  - PYTHON_VERSION=3.6
+  - PYTHON_VERSION=3.7
+  - PYTHON_VERSION=3.8
+before_install:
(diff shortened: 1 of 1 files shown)


jtpio:
Are there any reasons for not choosing an alternative like GitHub Actions when starting "fresh"? (just curious)

Maybe because of conda?

martinRenou:
No reason, honestly. I know Travis, I used it.

> Maybe because of conda?

What do you mean? Can you not use conda with Github Actions?

jtpio:
No worries, it was just tempting to ask since core ipywidgets moved to GH Actions.

> Can you not use conda with Github Actions?

It is possible (used for Voila), but it's not as nice and out-of-the-box as the `setup-python` start action for example.

martinRenou:
Ok. Yes, it might make sense to use Github Actions then.

PR #516, opened 2020-03-11
Add Image_slider.ipynb example, use setUrl and setBounds in imageOverlay

### LINKED ISSUE #514: Including the setUrl function to ImageOverlay
Hi, I'm trying to plot a sequence of images as a controlled animation. What I have tried is to link the "url" property of the ImageOVerlay instance to the "value" property of a play widget of ipywidgets. But this results in the flicker of the view like this:

Peek 2020-03-11 12-05

My guess of the reason behind is that every time the url is updated, a new "img" DOM is initialized instead of redrawing the old one. In this case, the [setUrl](https://leafletjs.com/reference-1.6.0.html#imageoverlay-seturl) function is more suitable. Is there any plan to include the setUrl function for ImageOverlay?

davidbrochart:
On `url` or `bounds` change, the layer was removed and added again, which caused flickering. We now use `setUrl` and `setBounds` on the `imageOverlay` object, which is much smoother.
The example shows how to use a slider to walk through a time series of images, which is similar to a video overlay but offers more control over the animation.
Fixes #514


js/src/layers/ImageOverlay.js
```diff
@@ -35,11 +35,7 @@ export class LeafletImageOverlayView extends rasterlayer.LeafletRasterLayerView
       'change:url',
       function() {
         const url = this.model.get('url');
-        const bounds = this.model.get('bounds');
-        const options = this.get_options();
-        this.map_view.obj.removeLayer(this.obj);
-        this.obj = L.imageOverlay(url, bounds, options);
-        this.map_view.obj.addLayer(this.obj);
+        this.obj.setUrl(url);
```
martinRenou: :+1: :+1: :+1: :+1:

Issue #454, opened 2020-01-04
Add dynamic styles to Choropleths?


deeplook:
At the moment a Choropleth accepts only one kind of data (in the `choro_data` parameter) to control the fill color of GeoJSON features. I figured it would be helpful to extend this further and control the entire styling of single features. I my case I'm most interested in setting line weights, but setting other style attributes would be also useful.

I've looked at [Leaflet's Choropleth example](https://leafletjs.com/examples/choropleth/) and tweaked it only a bit to generate the effect I would like to see in the image below.

<img width="667" alt="Screenshot 2020-01-04 at 12 41 04" src="https://user-images.githubusercontent.com/1001778/71765118-e0658600-2ef0-11ea-98a0-575d5f741325.png">

This was done by modifying the `style` function in a minimal way (the `weight` attribute):

```javascript
	function style(feature) {
		return {
			// weight: 2,
			weight: Math.random() * 10, 
			opacity: 1,
			color: 'white',
			dashArray: '3',
			fillOpacity: 0.7,
			fillColor: getColor(feature.properties.density)
		};
	}
```

Ideally one could implement this with a new dynamic `style` parameter replacing the existing `choro_data`, `colormap`, and `border_color`, which would break existing code. So maybe add such a more generic `style` parameter and respect (and deprecate) the previous parameters for some time before removing them? The new `style` parameter could be a mapping from `id` keys to style dicts and/or a function returning such style dicts.

I think this would add quite a few steroids to the existing Choropleths, and I'm glad to help with the implmentation. Opinions?

martinRenou:
This could be done easily actually. The Choropleth layer in ipyleaflet is just a GeoJSON layer, that we modify (fully in Python) before sending to the front-end: https://github.com/jupyter-widgets/ipyleaflet/blob/master/ipyleaflet/leaflet.py#L603. So it's "only" Python code, which should be easier to modify for newcomers.

omanges:
I would like to work on this, @martinRenou  please assign this to me.

martinRenou:
Woaw, thanks a lot!! Looking forward to this :)

omanges:
I think here user can just pass an entire dict of styles with key as the feature ID, which can be set to the style attribute of each Feature at this step https://github.com/jupyter-widgets/ipyleaflet/blob/9066d48edf31f917519b8c486f05061367c34890/ipyleaflet/leaflet.py#L618 correct me if I am wrong  @martinRenou :)

[...]

PR #518, opened 2020-03-16
Change made to add style_callback parameter for Choropleth map

omanges:
Made changes to add a new parameter named style_callback for the Choropleth map so that the user can pass a callback function for styling. Reference issue:- https://github.com/jupyter-widgets/ipyleaflet/issues/454


martinRenou:
Great!! Thanks a lot for the PR. I'll have a look at it.

omanges:
Hi @martinRenou, the pipeline issue are resolved now.

martinRenou:
Yes, thanks! I'll look at your PR today

omanges:
Actually the current changes I made are choropleth map, it would be more useful if this is made available asa GeoJson class trait ?

martinRenou:
True. But you don't have the colormap in the GeoJSON layer.

Actually both layers are pretty close. I am not really sure what's best to do. I think we wanted to do something close to what Folium does.

martinRenou:
See https://github.com/python-visualization/folium/blob/master/folium/features.py#L1022

omanges:
Actually, then we can allow the user to just specify a call_back function like following:
```
function style(feature) {
		return {
			// weight: 2,
			weight: Math.random() * 10, 
			opacity: 1,
			color: 'white',
			dashArray: '3',
			fillOpacity: 0.7,
			fillColor: getColor(feature.properties.density)
		};
	}
```
Like the original requirement, so this call_back is indepedent.

martinRenou:
Isn't it what you are currently doing?

omanges:
Yes, but I had made changes just for Choropleth Map, now I realized it can be made to the base class GeoJSON.

martinRenou:
So would the colormap be an attribute of the GeoJSON layer in your opinion?

omanges:
Actually we can add style_callback as an attribute to the GeoJSON and add _get_data function to update data.

martinRenou:
Not sure we want to get rid of the `Choropleth` layer though.. It would be nice to stay backward compatible

omanges:
No, I am not removing Choropleth, I will make changes and add it in this PR to make it more clear.

martinRenou:
Sure! Thanks a lot for looking into this. That's really helpful

omanges:
Hi @martinRenou , I have pushed the changes that I had suggesting please have a look.

ipyleaflet/leaflet.py
```diff
@@ -536,13 +536,34 @@ class GeoJSON(FeatureGroup):
     style = Dict().tag(sync=True)
     hover_style = Dict().tag(sync=True)
     point_style = Dict().tag(sync=True)
+    style_callback = Any()
```
martinRenou: We could put a custom validator on this trait (see [traitlets validation](https://traitlets.readthedocs.io/en/stable/using_traitlets.html#validation)). That would validate that it's a callable (functor/function/lambda) using [the built-in callable function](https://docs.python.org/3/library/functions.html#callable).

I'm fine if we do it later though.

ipyleaflet/leaflet.py
```diff
@@ -536,13 +536,34 @@ class GeoJSON(FeatureGroup):
...
 
     _click_callbacks = Instance(CallbackDispatcher, ())
     _hover_callbacks = Instance(CallbackDispatcher, ())
 
+    @observe('data', 'style', 'style_callback')
+    def _update_data(self, change):
```
martinRenou: It seems to shadow the `Choropleth._update_data` method. Maybe rename `Choropleth._update_data` to `Choropleth._update_choropleth_data` ?

examples/Choropleth.ipynb
```diff
@@ -51,6 +52,46 @@
...
+    "    }"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "m.remove_layer(layer)\n",
```
martinRenou: If you fix the shadowing of `Choropleth._update_data`, you should be able to dynamically change the `style_callback`. So no need to remove the layer and add it back.

You should be able to write:
```python
layer.style_callback = compute_style
```
And that would work.

ipyleaflet/leaflet.py
```diff
@@ -611,13 +631,18 @@ def _get_data(self):
...
         colormap = self.colormap.scale(self.value_min, self.value_max)
-        color_dict = {key: colormap(self.choro_data[key]) for key in self.choro_data.keys()}
-
         data = copy.deepcopy(self.geo_data)
-        for feature in data['features']:
-            feature['properties']['style'] = dict(fillColor=color_dict[feature['id']],
-                                                  color=self.border_color,
-                                                  weight=0.9)
+
+        if self.style_callback:
```
martinRenou: I wonder if it would not be cleaner to use a default value for `style_callback` in the case of the `Choropleth` class, something like:

```python
@default('style_callback')
def _default_style_callback(self):
    def compute_style(feature, colormap, choro_data):
        return dict(
            fillColor=colormap(choro_data),
            color='black',
            weight=0.9
        )
    return compute_style
```

And we could remove this `border_color` attribute if needed.

ipyleaflet/leaflet.py
omanges: I have added validation for style_callback. Thanks

ipyleaflet/leaflet.py
omanges: Update name to _update_choropleth_data

examples/Choropleth.ipynb
omanges: Removed the statement

ipyleaflet/leaflet.py
omanges: Added a default value for style_callback. Thanks

examples/Choropleth.ipynb
```diff
@@ -51,6 +52,45 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "layer_callback = ipyleaflet.Choropleth(\n",
+    "    geo_data=geo_json_data,\n",
+    "    choro_data=unemployment,\n",
+    "    colormap=linear.YlOrRd_04, \n",
+    "    border_color='black',\n",
+    "    style_callback=compute_style\n",
+    ")"
```
martinRenou: I meant here you can replace the entire code by: 
```python
layer.style_callback = compute_style
```

martinRenou:
Thanks a lot for your work!

examples/Choropleth.ipynb
omanges: Done :)

PR #530, opened 2020-03-25
Renamed positional argument for handle_draw()

giswqs:
Renamed the positional argument from `self` to `target` to avoid confusion.


martinRenou:
Thanks a lot!

Issue #446, opened 2019-11-21
Unreasonable requirement of 'id' property in GeoJSON for choropleth


andrhua:
Currently ipyleaflet obligates a geojson passed to `Choropleth` constructor to have 'id' property on each `feature` as mentioned in #361:
```
 {"type": "FeatureCollection",
"features":[ {"type":"Feature",
"id":"AL",
"properties":{"name":"Alabama"},
"geometry":{"type":"Polygon",
"coordinates": [[[-87.359296,35.00118]]]
```
I find this at least obscure, since now I have to waste my time assigning these ids when they are already present in `feature`, just going by the other name in its properties! Why can't we just use name provided by user himself as `id`?

martinRenou:
Hello @andrhua and thanks for opening an issue. Would you like to open a PR for fixing this? The `id` key is used here: https://github.com/jupyter-widgets/ipyleaflet/blob/master/ipyleaflet/leaflet.py#L578.

Adding a customizable `name` trait is indeed a good idea.

andrhua:
Hey @martinRenou, I'm surely gonna take a look, just need some time to read code carefully.

martinRenou:
I feel like we should follow what Folium does with the `key_on` parameter.

martinRenou:
@omanges would you like to give this one a try? As it is pretty related to what you've done.

omanges:
@martinRenou Yes, I can look into this. So here we need to first check whether the `name` is present and if not `id` is used. So either name or id is needed right?

martinRenou:
I would say we stick to the Folium API. Adding a `key_on` Unicode attribute, which would be `"id"` by default. And then users can set it to whatever they want :)

omanges:
Ohh cool :), you can assign this issue to me I can work on this.

martinRenou:
Thanks very much for your work! It's really appreciated :)

martinRenou:
Fixed by https://github.com/jupyter-widgets/ipyleaflet/pull/537

PR #537, opened 2020-03-29
To include a new trait key_on in Choropleth Map

omanges:
Added a new trait key_on in Choropleth Map as specified in #446


ipyleaflet/leaflet.py
```diff
@@ -637,6 +638,10 @@ def compute_style(feature, colormap, choro_data):
 
         return compute_style
 
+    @default('key_on')
+    def _default_key_on(self):
```
martinRenou: You actually don't need this method, I suggested doing it for the `style_callback` because it was cleaner in this special case, but for the `key_on` you can define it as `key_on = Unicode('id')`. It will then default to 'id'.

omanges: Ohh sure I'll do the changes.

martinRenou:
Amazing! Thanks a lot for this!

wcwou:
I'm new to ipyleaflet and am running into a problem.  Looking online, I see that this problem appears to have been raised and dealt with already (https://github.com/jupyter-widgets/ipyleaflet/issues/446),  but I am still encountering it.

The problem is that the ipyleaflet Choropleth automatically defaults to a key_on of 'id' even when one specifies something else.  For example, if one runs the example code at https://ipyleaflet.readthedocs.io/en/latest/api_reference/choropleth.html:

layer = ipyleaflet.Choropleth(
    geo_data=geo_json_data,
    choro_data=unemployment,
    colormap=linear.YlOrRd_04,
    border_color='black',
    style={'fillOpacity': 0.8, 'dashArray': '5, 5'})

Even if you add in "key_on = 'nonsense_word'", the Choropleth will still appear. 
 
Conversely, for my GeoJSON data, I need to specify the key as "properties.name", but when I do I get the following error (this is the last part of the trace):

~/anaconda3/lib/python3.7/site-packages/ipyleaflet/leaflet.py in _get_data(self)
    562         data = copy.deepcopy(self.geo_data)
    563         for feature in data['features']:
--> 564             feature['properties']['style'] = dict(fillColor=color_dict[feature['id']],
    565                                                   color=self.border_color,
    566                                                   weight=0.9)

KeyError: 'id'

It appears that the code is still looking for 'id' even though the key is specified as something else.

Issue #409, opened 2019-07-18
Choropleth layer with np.integer data


martinRenou:
As pointed out by @bill-coman:

`value_min` and `value_max` attributes being Float traits, passing a numpy array to `choro_data` containing integers can throw an exception.

PR #539, opened 2020-03-29
Fix Choropleth class when using integer data

martinRenou:
Should fix https://github.com/jupyter-widgets/ipyleaflet/issues/409


SylvainCorlay:
Great.
Do we want to do this with all `Float` traits of the package?

martinRenou:
I guess it would be harmless :) we can do it

PR #543, opened 2020-04-01
Vector Tile sample class to load Vector tile data

omanges:
Vector Tile sample class to load Vector tile data


ipyleaflet/leaflet.py
```diff
@@ -384,6 +388,30 @@ class Heatmap(RasterLayer):
     gradient = Dict({0.4: 'blue', 0.6: 'cyan', 0.7: 'lime', 0.8: 'yellow', 1.0: 'red'}).tag(sync=True, o=True)
 
 
+class VectorTileLayer(Layer):
+    _view_name = Unicode('LeafletVectorTileLayerView').tag(sync=True)
+    _model_name = Unicode('LeafletVectorTileLayerModel').tag(sync=True)
+
+    url = Unicode('https://vector.hereapi.com/v2/vectortiles/base/mc/{z}/{x}/{y}/omv?apiKey={apiKey}').tag(sync=True)
+    options = Dict({'apiKey': '0es5QrwgHuJCR5oNpzUHi_TiLyw-59jrC2JPiNfPwGE'})
```
martinRenou, line 396: If you want to add a new option, just do `api_key = Unicode('').tag(sync=True, o=True)`. ipyleaflet will automatically set the `apiKey` option.

PR #544, opened 2020-04-01
Vector Tile example changes

omanges:
Vector Tile example changes


ipyleaflet/leaflet.py
```diff
@@ -384,6 +388,28 @@ class Heatmap(RasterLayer):
     gradient = Dict({0.4: 'blue', 0.6: 'cyan', 0.7: 'lime', 0.8: 'yellow', 1.0: 'red'}).tag(sync=True, o=True)
 
 
+class VectorTileLayer(Layer):
+    _view_name = Unicode('LeafletVectorTileLayerView').tag(sync=True)
+    _model_name = Unicode('LeafletVectorTileLayerModel').tag(sync=True)
+
+    url = Unicode().tag(sync=True)
+    options = Dict().tag(sync=True)
```
martinRenou, line 396: This will conflict with the superclass's `options` attribute. You will need to remove it

js/src/layers/VectorTileLayer.js
```diff
@@ -0,0 +1,46 @@
...
+      url: '',
+      options: {}
+    };
+  }
+}
+
+export class LeafletVectorTileLayerView extends layer.LeafletLayerView {
+   create_obj() {
+        var options = {
+            ...this.model.get('options'),
```
martinRenou, line 21: and here use `this.get_options()`

ipyleaflet/leaflet.py
omanges: But as I need the options to be a Dict were as `Layer` class has options trait which is as `List`

martinRenou: The dict is computed for you in `this.get_options()` JavaScript side.

omanges: Ok I get it now

js/src/layers/VectorTileLayer.js
```diff
@@ -0,0 +1,45 @@
...
+      },
+      this
+    );
+  }
+
+  handle_message(content) {
+    if (content.msg == 'redraw') {
+      this.obj.redraw();
+    }
+  }
```
martinRenou, line 44: You also need to listen to options changes, and apply them. In WMSLayer there is a `setParam` method, I guess you have something equivalent in the Leaflet.VectorGrid.Protobuf

ipyleaflet/leaflet.py
```diff
@@ -384,6 +388,27 @@ class Heatmap(RasterLayer):
     gradient = Dict({0.4: 'blue', 0.6: 'cyan', 0.7: 'lime', 0.8: 'yellow', 1.0: 'red'}).tag(sync=True, o=True)
 
 
+class VectorTileLayer(Layer):
+    _view_name = Unicode('LeafletVectorTileLayerView').tag(sync=True)
+    _model_name = Unicode('LeafletVectorTileLayerModel').tag(sync=True)
+
+    url = Unicode().tag(sync=True)
```
martinRenou, line 395: You are also missing the `vector_tile_layer_styles` option from the Protobuf class.

js/src/layers/VectorTileLayer.js
omanges: Actually, as I see there is code there no way to update options and do a soft refresh in Leaflet.VectorGrid.Protobuf:- https://github.com/Leaflet/Leaflet.VectorGrid/issues/159

ipyleaflet/leaflet.py
omanges: Added thank you

js/src/layers/VectorTileLayer.js
martinRenou: I see. I guess we can try using `setOptions`. But let's not block this PR any longer.

PR #552, opened 2020-04-21
Use leaflet-defaulticon-compatibility for icon image bundling

### LINKED ISSUE #551: Marker does not show in the jupyterlab with version 0.12.4
Marker does not show in the jupyterlab with version 0.12.4

result: `get http://127.0.0.1:8888/[object%20Module] 404`

davidbrochart:
Fixes #551


SylvainCorlay:
This is great. I left a comment in the repository for this package asking if they would consider upstreaming the changes...

PR #553, opened 2020-04-23
Fix setup call in setup.py

martinRenou:
This way Github can more easily parse the `setup.py` file and find the package name. This will hopefully make Github see ipyleaflet as a Python package and fix the "used by" button.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -126,33 +126,33 @@ version_ns = {}
 with open(os.path.join(here, 'ipyleaflet', '_version.py')) as f:
     exec(f.read(), {}, version_ns)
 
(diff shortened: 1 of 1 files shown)


martinRenou:
Although this is what Github has reported me, I don't think this is the cause of the issue. It works just fine in `ipycanvas` (where I do the same with `setup_args`). I actually think this is coming from the fact that there is more JavaScript code than there is Python code. Same for Voila.

I see there is a way to tweak Github https://github.community/t5/How-to-use-Git-and-GitHub/how-to-change-the-category/td-p/27790. Let's see if that works.

Issue #555, opened 2020-04-23
Problem with GeoJSON style using version 0.12.4


vincentchabot:
Hi, I get a problem with the latest version of ipyleaflet when using a geojson file containg style properties. 

```python
import ipyleaflet as ipyl 
import json 
with open("my.json") as json_file:
    geo_contour = json.load(json_file)
# See the properties of geojson files for a given contour 
print(geo_contour["features"][0]["properties"])
my_map = ipyl.Map(center=(45,4.5),zoom=8,layout={"height":"500px"})
layer = ipyl.GeoJSON(data=geo_contour,hover_style={"opacity":1},name="AROME")
my_map.add_layer(layer)
# See the properties of geojson files for a given contour 
print(geo_contour["features"][0]["properties"])
# Display the map 
my_map
```

# Output with ipyleaflet 0.12.4 
### First print 
```python 
{'value': 4.0,
 'units': None,
 'name': 'unknown',
 'cmap': {'value': 4.0, 'mini': 4.0, 'maxi': 84.0},
 'style': {'fillColor': '#67001f',
  'fillOpacity': 0.9,
  'opacity': 0.2,
  'color': '#000000',
  'dashArray': '5',
  'weight': 2}}
```
### Second print 
We can see that geo_contour has been modified by GeoJSON (style dictionnary is now empty). 

```python 
{'value': 4.0,
 'units': None,
 'name': 'unknown',
 'cmap': {'value': 4.0, 'mini': 4.0, 'maxi': 84.0},
 'style': {}}
```
### Map 
![map_0 12 4](https://user-images.githubusercontent.com/33070178/80080998-a6c59300-8552-11ea-8632-acdb2bb470e4.png)

# Output  with ipyleaflet 0.12.3
### First print 
```python 
{'value': 4.0,
 'units': None,
 'name': 'unknown',
 'cmap': {'value': 4.0, 'mini': 4.0, 'maxi': 84.0},
 'style': {'fillColor': '#67001f',
  'fillOpacity': 0.9,
  'opacity': 0.2,
  'color': '#000000',
  'dashArray': '5',
  'weight': 2}}
```
### Second print 
```python 
{'value': 4.0,
 'units': None,
 'name': 'unknown',
 'cmap': {'value': 4.0, 'mini': 4.0, 'maxi': 84.0},
 'style': {'fillColor': '#67001f',
  'fillOpacity': 0.9,
  'opacity': 0.2,
  'color': '#000000',
[...]

PR #559, opened 2020-04-25
Added changes to avoid setting empty style for GeoJson object 

omanges:
This is a fix for the bug reported in https://github.com/jupyter-widgets/ipyleaflet/issues/555


martinRenou:
Thanks a lot for opening a PR. I am not forgetting about this, we need to merge it before making a release.

ipyleaflet/leaflet.py
```diff
@@ -579,7 +579,7 @@ def _get_data(self):
             elif self.data['type'] == 'FeatureCollection':
                 for feature in self.data['features']:
                     feature['properties']['style'] = self.style_callback(feature)
-        else:
+        elif self.style:
             if self.data['type'] == 'Feature':
                 self.data['properties']['style'] = self.style
```
martinRenou, line 584: In the method, we should replace all the `self.data['properties']['style'] = something` into `self.data['properties']['style'].update(something)`

martinRenou: this way we never overwrite the style, we simply update it.

omanges: But if the style key is not present within the `feature` then will it give a `KeyError`

omanges: Made changes as suggested :)

ipyleaflet/leaflet.py
```diff
@@ -571,16 +571,16 @@ def _update_data(self, change):
...
-        else:
+                    feature['properties']['style'].update(self.style_callback(feature))
+        elif self.style:
             if self.data['type'] == 'Feature':
-                self.data['properties']['style'] = self.style
+                self.data['properties']['style'].update(self.style)
             elif self.data['type'] == 'FeatureCollection':
                 for feature in self.data['features']:
-                    feature['properties']['style'] = self.style
+                    feature['properties']['style'].update(self.style)
```
sackh, line 583: There should be a minor change here, as this code always expects that there will be `style` in properties of `geojson`. If `style` is not present in the properties this fails with `KeyError` exception. I think there should be one more check if `style` key is already present then update, else assign the value.
e.g. - 
```python
if 'style' in  feature['properties']:
   feature['properties']['style'].update(self.style)
else:
   feature['properties']['style'] = self.style
```

sackh: for all update statements, need to check if `style` key is already present or not.

sackh: I have raised PR #572

PR #561, opened 2020-04-27
Remove xarray dependency 

stephaniegott:
Removes `xarray` from the dependencies of ipyleaflet.

Although ipyleaflet has `xarray` listed in its `install_requires` I'm not exactly sure why it needs this dependency. It would be great if when I installed ipyleaflet, I didn't also have to install `xarray` (which, in turn, pulls in `pandas`) as a transitive dependency.

### DIFF
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -334,8 +334,9 @@ class Velocity(Layer):
     except ModuleNotFoundError:
         raise ImportError(
             'The `xarray` module (required for use of the Velocity layer) is'
-            'not installed. Run "$ pip install xarray" to install the latest'
-            'version.'
+            'not installed. Run "$ conda install -c conda-forge xarray dask'
+            'netCDF4 bottleneck" or "$ pip install xarray" to install the '
+            'latest version.'
         )
 
     _view_name = Unicode('LeafletVelocityView').tag(sync=True)
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -329,6 +329,15 @@ class VideoOverlay(RasterLayer):
 
 
 class Velocity(Layer):
+    try:
+        import xarray
+    except ModuleNotFoundError:
+        raise ImportError(
+            'The `xarray` module (required for use of the Velocity layer) is'
+            'not installed. Run "$ pip install xarray" to install the latest'
+            'version.'
+        )
+
     _view_name = Unicode('LeafletVelocityView').tag(sync=True)
     _model_name = Unicode('LeafletVelocityModel').tag(sync=True)
 
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -334,9 +334,7 @@ class Velocity(Layer):
     except ModuleNotFoundError:
         raise ImportError(
             'The `xarray` module (required for use of the Velocity layer) is'
-            'not installed. Run "$ conda install -c conda-forge xarray dask'
-            'netCDF4 bottleneck" or "$ pip install xarray" to install the '
-            'latest version.'
+            'not installed.'
         )
 
     _view_name = Unicode('LeafletVelocityView').tag(sync=True)
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -333,7 +333,7 @@ class Velocity(Layer):
         import xarray
     except ModuleNotFoundError:
         raise ImportError(
-            'The `xarray` module (required for use of the Velocity layer) is'
+            'The `xarray` module (required for use of the Velocity layer) is '
             'not installed.'
         )
 
--- a/setup.py
+++ b/setup.py
@@ -137,8 +137,7 @@ setup_args = {
     'install_requires': [
         'ipywidgets>=7.5.0,<8',
         'traittypes>=0.2.1,<3',
-        'xarray>=0.10',
-        'branca>=0.3.1,<0.4'
+        'branca>=0.3.1,<0.4',
     ],
     'packages': find_packages(),
     'zip_safe': False,


davidbrochart:
`xarray` is needed for the Velocity layer. With some work, we could probably do with just `numpy`. Or not include `xarray` in the requirements, and throw an error saying that `xarray` is needed if a Velocity layer is used.

martinRenou:
> With some work, we could probably do with just `numpy`

I'd be in favor with this :+1:. In the end the Velocity layer only needs a grid of velocity data, which can definitely be expressed by a NumPy array.

This would:
- make the Velocity layer more general and simple (as people will be able to pass a NumPy array built from any kind of data structure: Pandas, xarray). And it would no longer be required to pass an xarray that has a very specific structure.
- remove the xarray dependency which, if I remember correctly, is quite heavy
- remove some code complexity from ipyleaflet

stephaniegott:
@davidbrochart @martinRenou Thanks for the feedback. Subbing in `numpy` for `xarray` sounds like a good plan of action. In the meantime, could we perhaps add an `extras_require` for Velocity? Something like:
```
setup_args = {
    'name':'ipyleaflet',
    ...
    'extras_require': {
        'velocity':  ['xarray>=0.10'],
    }
)
```

davidbrochart:
I'm not sure it is such a good idea to try and implement ourselves what xarray offers. Dimension-based indexing is really powerful. Say you only want to show a subset of the data in ipyleaflet, you can select the lat/lon range xarray. If it's big data, it can be backed by Zarr and Dask so you load chunks of data. Do we really want to give up on that?
Actually I think the second option is the best, only import xarray in Velocity, and remove it from the requirements altogether.

martinRenou:
I am not saying we reimplement what xarray offers.

> Say you only want to show a subset of the data in ipyleaflet, you can select the lat/lon range xarray. If it's big data, it can be backed by Zarr and Dask so you load chunks of data. Do we really want to give up on that?

I actually don't see why it would be "giving up on that". The user could still perform whatever they wants with their data, selecting/filtering with Pandas or xarray or another library, but they would need to turn it into a simple data structure like a NumPy array before passing it to the `VelocityLayer`.
This way we don't make the choice for the user of what kind of library can be used, and we make our API simpler.

davidbrochart:
I don't know, I also see some nice potential ipyleaflet/xarray integration. It's not implemented yet, but we could imagine ipyleaflet displaying only the lat/lon box that is currently shown, and querying the data it needs through the `xarray.Dataset`, that could trigger reading from disk or remote network... It would be similar to Datashader in that regard.

martinRenou:
Well that is technically already possible writing some Python code. And it sounds out of scope to me for a Python port of LeafletJS for Jupyter.

davidbrochart:
I don't know if it's out of scope, we may also want to leverage the Python ecosystem and allow the user to do powerful things while writing less code.

martinRenou:
But that sounds really like ipyleaflet taking another route, that is what I mean. Maybe it would be better to make your idea another library.

Otherwise I fear ipyleaflet might become a huge beast with lots of dependencies that tries to fix the world's problems, but in the end doesn't do it well.
IMO it's better to have a narrow scope for libraries. A good library, in my definition, is a library that doesn't do much, but does it well.

davidbrochart:
Yes, you are right. I think an xarray/leaflet integration deserves a separate package. This is something that has been mentioned by the xarray community (see https://github.com/benbovy/xvelmap/issues/5#issuecomment-396852106). I am going to start `xarray-leaflet` for that purpose!

stephaniegott:
While the xarray/leaflet integration is in progress it seems like there are two options for removing the dependency on `xarray`: 
1. Use `extras_require` so that users need to install `ipyleaflet[velocity]` if they want to use the Velocity layer. 
2. Only import `xarray` in velocity (as @davidbrochart suggested) and throw an error there if users don't have it installed. 
Do either of these sound like something you folks would be okay with for right now?

davidbrochart:
I prefer option 2. And option 1 would work with pip but not conda.

martinRenou:
Good, should we merge this? We need to remember to remove the xarray dependency in the conda package as well when doing the release.

ipyleaflet/leaflet.py
```diff
@@ -329,6 +329,15 @@ class VideoOverlay(RasterLayer):
 
 
 class Velocity(Layer):
+    try:
+        import xarray
+    except ModuleNotFoundError:
+        raise ImportError(
+            'The `xarray` module (required for use of the Velocity layer) is'
+            'not installed. Run "$ pip install xarray" to install the latest'
+            'version.'
```
davidbrochart, line 338: Should we tell how to install? pip is not the only way.

stephaniegott: I added the instruction for installing with conda as well.

davidbrochart: I meant, maybe we should not say how to install. Sorry if it was not clear.

stephaniegott: Oh I see. I do think having the instructions might be nice for someone new to these packages but I could remove them to make the message simpler.

davidbrochart: I'd rather not have them. @martinRenou what do you think?
(and you don't need dask, netCDF4 or bottleneck in order to install xarray)

martinRenou: I agree. Although for beginners it's nice to have an install instruction. But we should not take any guess on what kind of tool they use to install their packages, and telling someone to install something with `pip` when they actually use `conda` might break their environment if they are not careful.

ipyleaflet/leaflet.py
```diff
@@ -334,9 +334,7 @@ class Velocity(Layer):
     except ModuleNotFoundError:
         raise ImportError(
             'The `xarray` module (required for use of the Velocity layer) is'
```
davidbrochart, line 336: Don't we need a space after "is"?

davidbrochart:
LGTM. Thanks @stephaniegott !

martinRenou:
Thanks a lot!

jasongrout:
Corresponding conda-forge issue at https://github.com/conda-forge/ipyleaflet-feedstock/issues/80

martinRenou:
Great thanks a lot Jason :)

PR #568, opened 2020-04-30
Fix VectorTile layer

martinRenou:
The VectorTile layer is actually not working in the latest release (my fault :confused:), the constructor was calling a non-existing method

@omanges


omanges:
Thank you for the fix 😁

PR #573, opened 2020-05-02
Refactor GeoJSON._get_data

martinRenou:
@sackh @omanges If you could tell me what you think and give it a try that would be super great. On my side it seems everything is working fine as expected, as well as the regression you were fixing @sackh


sackh:
@martinRenou - I will quickly, try these changes at my end. I will let you know.

sackh:
@martinRenou - changes look good to me, I tried all cases at my end and those are working fine for me. 💯

Issue #534, opened 2020-03-28
Search Control


SylvainCorlay:
We should add support for the [search control](https://github.com/stefanocudini/leaflet-search) in ipyleaflet.

This is use in a couple of applications like this map here, on the natural radioactivity in France https://cartoradon.irsn.fr/static/

SylvainCorlay:
```
  var search_url = '/commune.py/communes/search/'+region+'/{s}?';
	var searchControl = new L.Control.Search({
			url: search_url,
			jsonpParam: 'json_callback',
			propertyName: 'display_name',
			propertyLoc: ['lat','lon'],
			//markerLocation: true,
			autoType: false,
			autoCollapse: false,
			minLength: 1,
         zoom:10,
         animateLocation:false,
         position:'topright'
      });
   map.addControl(searchControl);
   searchControl.expand();

   function onMapClick(e) {
      search_by_click(e.latlng);
   }
   map.on('click', onMapClick);
```

giswqs:
This would be a very nice addition. The Search Control can search both place names and maker properties. [Here](https://labs.easyblog.it/maps/leaflet-search/examples/geocoding-nominatim.html) is another example using the Search Control.

SylvainCorlay:
@giswqs do you want to take a stab at it?

giswqs:
@SylvainCorlay I am not good at JavaScript, and I am still learning the ipyleaflet source code. It might take some time before I can contribute to the source code at this level. Could you elaborate a little bit more about the general workflow for adding a new control to ipyleaflet?

Using the `ZoomControl` as an example, it seems these are the steps: 

1. Add a `SearchControl.js` to [ipyleaflet/js/src/controls/](https://github.com/jupyter-widgets/ipyleaflet/tree/master/js/src/controls)
2. Modify [jupyter-leaflet.js](https://github.com/jupyter-widgets/ipyleaflet/blob/master/js/src/jupyter-leaflet.js) to include the SearchControl.
3. Modify [leaflet.py](https://github.com/jupyter-widgets/ipyleaflet/blob/master/ipyleaflet/leaflet.py) to include the SearchControl
4. Add a notebook example to the [examples](https://github.com/jupyter-widgets/ipyleaflet/tree/master/examples) folder. 

But I don't know how to incorporate the Search Control source code from [[leaflet-search/src/](https://github.com/stefanocudini/leaflet-search/tree/master/src)] into ipyleaflet.

[...]

PR #576, opened 2020-05-05
#534 - Code changes for Search Control feature

sackh:
Hi @martinRenou,
I have added changes for SearchControl implementation.
Please review.

### DIFF
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -989,7 +989,7 @@ class SearchControl(Control):
     auto_type = Bool(False).tag(sync=True, o=True)
     auto_collapse = Bool(False).tag(sync=True, o=True)
     animate_location = Bool(False).tag(sync=True, o=True)
-    marker = Instance(Marker).tag(sync=True, **widget_serialization)
+    marker = Instance(Marker).tag(sync=True, o=True, **widget_serialization)
 
 
 class MapStyle(Style, Widget):
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -983,9 +983,13 @@ class SearchControl(Control):
     _model_name = Unicode('LeafletSearchControlModel').tag(sync=True)
     url = Unicode().tag(sync=True, o=True)
     zoom = Int(10).tag(sync=True, o=True)
-    propertyName = Unicode('display_name').tag(sync=True, o=True)
-    propertyLoc = List(['lat', 'lon']).tag(sync=True, o=True)
-    jsonpParam = Unicode('json_callback').tag(sync=True, o=True)
+    property_name = Unicode('display_name').tag(sync=True, o=True)
+    property_loc = List(['lat', 'lon']).tag(sync=True, o=True)
+    jsonp_param = Unicode('json_callback').tag(sync=True, o=True)
+    auto_type = Bool(False).tag(sync=True, o=True)
+    auto_collapse = Bool(False).tag(sync=True, o=True)
+    animate_location = Bool(False).tag(sync=True, o=True)
+    marker = Instance(Marker).tag(sync=True, **widget_serialization)
 
 
 class MapStyle(Style, Widget):
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -978,6 +978,16 @@ class LegendControl(Control):
         self.send_state()
 
 
+class SearchControl(Control):
+    _view_name = Unicode('LeafletSearchControlView').tag(sync=True)
+    _model_name = Unicode('LeafletSearchControlModel').tag(sync=True)
+    url = Unicode().tag(sync=True, o=True)
+    zoom = Int(10).tag(sync=True, o=True)
+    propertyName = Unicode('display_name').tag(sync=True, o=True)
+    propertyLoc = List(['lat', 'lon']).tag(sync=True, o=True)
+    jsonpParam = Unicode('json_callback').tag(sync=True, o=True)
+
+
 class MapStyle(Style, Widget):
     """ Map Style Widget """
     _model_name = Unicode('LeafletMapStyleModel').tag(sync=True)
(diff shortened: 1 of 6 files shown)


ipyleaflet/leaflet.py
```diff
@@ -978,6 +978,16 @@ def remove_legend_element(self, key):
...
 
 
+class SearchControl(Control):
+    _view_name = Unicode('LeafletSearchControlView').tag(sync=True)
+    _model_name = Unicode('LeafletSearchControlModel').tag(sync=True)
+    url = Unicode().tag(sync=True, o=True)
+    zoom = Int(10).tag(sync=True, o=True)
+    propertyName = Unicode('display_name').tag(sync=True, o=True)
+    propertyLoc = List(['lat', 'lon']).tag(sync=True, o=True)
+    jsonpParam = Unicode('json_callback').tag(sync=True, o=True)
```
martinRenou, line 988: Attribute names should be in `snake_case` Python side, we have some logic to turn them automatically to `camelCase` on the JavaScript side.

js/src/controls/SearchControl.js
```diff
@@ -0,0 +1,49 @@
...
+const L = require('../leaflet.js');
+const control = require('./Control.js');
+
+export class LeafletSearchControlModel extends control.LeafletControlModel {
+  defaults(){
+    return{
+      _view_name: 'LeafletSearchControlView',
+      _model_name: 'LeafletSearchControlModel',
+      url: null,
+      jsonpParam: 'json_callback',
```
martinRenou, line 13: Same here, they should reflect the Python attribute names here.

js/src/controls/SearchControl.js
```diff
@@ -0,0 +1,49 @@
...
+      propertyName: 'display_name',
+      propertyLoc: ['lat','lon'],
+      autoType: false,
```
martinRenou, line 16: This attribute and some others are missing Python side

js/src/controls/SearchControl.js
```diff
@@ -0,0 +1,49 @@
...
+  initialize(parameters) {
+    super.initialize(parameters);
+    this.map_view = this.options.map_view;
+  }
+
+  create_obj() {
+    this.obj = L.control.search(this.get_options());
+  }
+
+  get_options() {
```
martinRenou, line 36: You should not need this `get_options`. You can safely remove it once you turned all the attributes in `snake_case`. The `get_options` method is already implemented here: https://github.com/jupyter-widgets/ipyleaflet/blob/3b767b0f4895127c707bee1ee43d089f6f55529b/js/src/utils.js#L17

ipyleaflet/leaflet.py
sackh: Done.

js/src/controls/SearchControl.js
sackh: Done.

js/src/controls/SearchControl.js
```diff
@@ -10,13 +10,13 @@ export class LeafletSearchControlModel extends control.LeafletControlModel {
...
-      autoCollapse: false,
+      jsonp_param: 'json_callback',
+      property_name: 'display_name',
+      property_loc: ['lat','lon'],
+      auto_type: false,
+      auto_collapse: false,
       zoom:10,
-      animateLocation:false,
+      animate_location:false,
       marker: L.circleMarker([0,0],{radius:30}),
```
martinRenou, line 20: This should be `null` by default.

You also need to specify a special widget serializer/deserializer for the `marker` attribute. You can look at the `Popup` implementation for an example: https://github.com/jupyter-widgets/ipyleaflet/blob/master/js/src/layers/Popup.js#L29.

js/src/controls/SearchControl.js
sackh: I have added them now.

js/src/controls/SearchControl.js
sackh: I have removed this now.

js/src/controls/SearchControl.js
```diff
@@ -32,18 +32,4 @@ export class LeafletSearchControlView extends control.LeafletControlView{
   create_obj() {
     this.obj = L.control.search(this.get_options());
```
martinRenou, line 33: You did not tell to ipyleaflet that the `marker` attribute was an option Python side (the tag `o=True` is missing), but I guess this is fine. Only the marker won't be part of the `get_options` output here, so you need to integrate it in some way. Something like:

```javascript
this.create_child_view(this.model.get('marker')).then((view) => {
  let options = this.get_options();
  options.marker = view;
  this.obj = L.control.search(options);
});
```

martinRenou: Again you can check the Popup implementation for the child view creation.

js/src/controls/SearchControl.js
sackh: Done.

js/src/controls/SearchControl.js
sackh: Done.

sackh: I have removed child view creation logic as it throws below exception.
```
leaflet-src.js:4756 Uncaught (in promise) TypeError: Cannot read property 'addTo' of undefined
    at e.addControl (leaflet-src.js:4756)
    at Map.js:160
    at async Promise.all (index 2)
```
It works fine without a child view for the marker. let me know your thoughts.

martinRenou: Yes, I feel like it's because you need to return the promise in the method, something like:

```javascript
return this.create_child_view(this.model.get('marker')).then((view) => {
  let options = this.get_options();
  options.marker = view;
  this.obj = L.control.search(options);
});
```

I'll test and merge your PR tomorrow or this week-end :)

sackh: Thanks.

martinRenou:
Let me push one final fix on your PR and that should be good to go :)

The marker should actually be `view.obj` in the code:
```python
return this.create_child_view(this.model.get('marker')).then((view) => {
  let options = this.get_options();
  options.marker = view.obj;
  this.obj = L.control.search(options);
});
```

This way you can now assign any marker, and if the marker changes Python side it will update accordingly:
search

Let me push the change and merge the PR. It should not require any action from your side.

martinRenou:
Actually I don't seem to have the right to push to your repo. Could you please make the change? :)

sackh:
Sure, I will make this change. Thanks for your help.

PR #577, opened 2020-05-07
Fix overwriting of colormap ranges for Choropleth

### LINKED ISSUE #570: Choropleth : value_min, value_max argument are not taken into account when creating it
When defining a choropleth (version 0.12.3 of ipyleaflet), value_max and value_min are compute from data even if keys are supplied. 

You can find below two "screenshot" of the Choropleth example notebook (provided in the doc). 
In the first one, "value_max" is set to 55.0 using the key. 
However, the layer.value_max is 10.3. 
Value_max1

When setting value_max to 55.0 after the creation of the Choropleth (second screenshot), the argument is taken into account (and it modify the layer appearance)
Expected

wendli01:
Fixes #570 by removing `_update_bounds`.

### DIFF
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -658,11 +658,6 @@ class Choropleth(GeoJSON):
     colormap = Instance(ColorMap)
     key_on = Unicode('id')
 
-    @observe('choro_data')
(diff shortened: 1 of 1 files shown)


martinRenou:
Thanks a lot!

PR #587, opened 2020-05-11
Add window_url attribute to Map

davidbrochart:
It can be useful to retrieve the URL of the browser window when we want to serve local tiles with `LocalTileLayer` through a custom Jupyter server handler.

### DIFF
--- a/js/src/Map.js
+++ b/js/src/Map.js
@@ -34,7 +34,6 @@ export class LeafletMapModel extends widgets.DOMWidgetModel {
       _model_module: 'jupyter-leaflet',
       _view_module: 'jupyter-leaflet',
 
-      window_url: '',
       center: DEFAULT_LOCATION,
       zoom_start: 12,
       zoom: 12,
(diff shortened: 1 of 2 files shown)


davidbrochart:
Ideally, I would like the window URL to be retrieved only once, but I don't quite know where it should be. Right now it will be retrieved each time the map bounds are changed.

js/src/Map.js
```diff
@@ -107,6 +108,7 @@ export class LeafletMapModel extends widgets.DOMWidgetModel {
       this.set('south', bounds.south);
       this.set('east', bounds.east);
       this.set('west', bounds.west);
+      this.set('window_url', window.location.href);
```
martinRenou, line 111: You can put this line where the map is initialized?

davidbrochart: You mean in the `initialize` method of the view?

martinRenou: It would be even better in the initialize method of the model, then it is called only once.

davidbrochart: Thanks.

js/src/Map.js
```diff
@@ -34,7 +34,6 @@ export class LeafletMapModel extends widgets.DOMWidgetModel {
       _model_module: 'jupyter-leaflet',
       _view_module: 'jupyter-leaflet',
 
-      window_url: '',
```
martinRenou: Shouldn't we keep this?

davidbrochart: It doesn't seem to be needed.

martinRenou: It is needed when having an embedded widget, as it retrieves the model from here. Anyway as we discussed this PR won't make sense in the case of an embedded widget.

PR #595, opened 2020-05-14
Add ipygee as related projects

fitoprincipe:
Hi =) Since I've contributed on this package in the past (95ef9d685f294e7f362e7d176adbc35d80219a66, 12782ce802761284b60e90a99fc5fddd30563028...), I think it will be fair to include my package in the related projects (although there is a very similar package already listed). Thank you


martinRenou:
Sure! Thanks for listing your package :)

fitoprincipe:
As always, thank you @martinRenou! Stay safe =)

martinRenou:
You too! :)

Issue #70, opened 2017-07-28
Feature Request: Add support for projections not supported by Leaflet


tylere:
The [Proj4Leaflet](http://kartena.github.io/Proj4Leaflet/) Javascript library adds support for additional projections in Leaflet, beyond the [standard projections](http://leafletjs.com/reference-1.1.0.html#projection). Could this also be added to IPyLeaflet so that non-standard projections could be requested? For example:

```
map = ipyleafletee.Map(
  crs='EPSG:3020',  # http://spatialreference.org/ref/epsg/wgs-84-antarctic-polar-stereographic/
  zoom=6,
  center=[0, 0],
  default_tiles=ipyleaflet.TileLayer(url=SOME_URL_THAT_RETURNS_PROJECTED_TILES)
)
map
```

fabricebrito:
+1

Juanlu001:
How to use the standard projections to begin with? https://leafletjs.com/reference-1.1.0.html#projection

betolink:
I was going to open this issue but it's already open so, is there any reason why this is not supported? 
I think a lot of people work with non standard projections and it would be nice-to-have feature. I can work on it and send a PR for next week.

davidbrochart:
There is no particular reason not to support it, and it would be great to have. I investigated the use of Proj4Leaflet with vector tiles (that we now have in ipyleaflet), which would have been really nice, but it doesn't seem to be possible (https://github.com/kartena/Proj4Leaflet/issues/173).
If you want to work on that, that would be awesome.

betolink:
I guess we can resolve this issue on the next release =)

davidbrochart:
Let's close it now, thank you!

martinRenou:
Thanks again :)

PR #598, opened 2020-05-15
Added support for non standard projections for Map and TileLayers

betolink:
This addresses #70, now we can use predefined polar projections or define our own projections. There is a notebook with an example of how to use this feature, the default is still 3857. 
Now we can define our map as follows:

```python
from ipyleaflet import (
    Map,
    projections
)
# note that we need to use the same projection for the our layer and the map.
m1 = Map(center=(90, 0),
        zoom=0,
        basemap=basemaps.NASAGIBS.BlueMarble3413,
        crs=projections.EPSG3413)
```

or we can define a custom projection with a dictionary:

```python
my_projection = {
    'name': 'EPSG:2163',
    'custom': True, #This is important, it tells ipyleaflet that this projection is not part of the predefined ones.
    'proj4def': '+proj=laea +lat_0=45 +lon_0=-100 +x_0=0 +y_0=0 +a=6370997 +b=6370997 +units=m +no_defs',
    'origin': [-2215235.82, 420644.41],
    'bounds': [
        [-8046094.81, 1819060.18],
        [988364.71, 3511186.72]   
    ],
    'resolutions': [
        8192.0,
        4096.0,
        2048.0,
        1024.0,
        512.0,
        256.0
    ]
}


m2 = Map(center=(40, -104),
        zoom=0,
        layers=(SOME_WMS_LAYER, ),
        crs=my_projection)
```

Another extra for this PR is to have a CHANGELOG file to keep track of changes and not just use the releasing notes. It would be cool if this projects implements some sort of semver and rendering tests.


davidbrochart:
That looks awesome @betolink!

betolink:
Thanks @davidbrochart! I'm still cleaning the code but should be ready to merge today.

davidbrochart:
Take your time! You might not get a review this week-end anyway ;-)

CHANGELOG.md
```diff
@@ -0,0 +1,55 @@
+## UNRELEASED
```
martinRenou, line 1: :clap: that's awesome. Thanks for doing this. We need a real changelog indeed.

ipyleaflet/leaflet.py
```diff
@@ -1038,8 +1039,8 @@ def _default_layers(self):
 
         return (basemap,)
 
-    bounds = Tuple(read_only=True)
-    bounds_polygon = Tuple(read_only=True)
+    bounds = Tuple(read_only=False)
+    bounds_polygon = Tuple(read_only=False)
```
martinRenou, line 1043: You can't change this, those two parameters needs to remain `read_only`

ipyleaflet/projections.py
```diff
@@ -0,0 +1,68 @@
+from traitlets.utils.bunch import Bunch
+
+projections = Bunch(
```
martinRenou, line 3: Awesome, I love it

js/src/Map.js
```diff
@@ -33,8 +36,8 @@ export class LeafletMapModel extends widgets.DOMWidgetModel {
       _model_name: 'LeafletMapModel',
       _model_module: 'jupyter-leaflet',
       _view_module: 'jupyter-leaflet',
-
       center: DEFAULT_LOCATION,
+      continuousWorld: false,
```
martinRenou, line 40: Why did you add this?

If you want to add a new option, it must be written in `snake_case`

ipyleaflet/leaflet.py
```diff
@@ -969,7 +970,7 @@ class Map(DOMWidget, InteractMixin):
     max_zoom = CFloat(18).tag(sync=True, o=True)
     min_zoom = CFloat(1).tag(sync=True, o=True)
     interpolation = Unicode('bilinear').tag(sync=True, o=True)
-    crs = Enum(values=allowed_crs, default_value='EPSG3857').tag(sync=True)
+    crs = Dict(default_value=projections.EPSG3857).tag(sync=True, o=True)
```
martinRenou, line 973: You should not add  the `o=True`, because the `crs` parameter is treated differently than other options JavaScript side.

js/src/layers/WMSLayer.js
```diff
@@ -18,6 +19,7 @@ export class LeafletWMSLayerModel extends tilelayer.LeafletTileLayerModel {
       transparent: false,
       version: '1.1.1',
       crs: null,
+      noWrap: true,
```
martinRenou, line 22: If you want to add a new option, it must be written in `snake_case`

js/src/leaflet.js
```diff
@@ -62,4 +64,17 @@ L.tileLayer.wms = function(url, options) {
...
+//         if (this.infinite) { return null; }
+
+//         var b = this.projection.bounds, // <-- projection does not have bounds property
+//                 // Thus it generates Uncaught TypeError: Cannot read property 'min' of undefined
+//             s = this.scale(zoom),
+//             min = this.transformation.transform(b.min, s),
+//             max = this.transformation.transform(b.max, s);
+
+//         return L.bounds(min, max);
+//     };
```
martinRenou, line 77: Maybe this is a leftover?

js/src/Map.js
davidbrochart: Also, we might want to make it configurable from the python side. Some projections may need the tile coordinates to be wrapped?
I can see that this option [used to exist](https://leafletjs.com/reference-0.7.7.html#tilelayer-continuousworld) but [doesn't seem to be present anymore](https://leafletjs.com/reference-1.0.3.html) in Leaflet 1.0.3, which Proj4Leaflet 1.0.1 uses.

js/src/layers/WMSLayer.js
davidbrochart: Some map CRS might wrap around, should it be configurable from python?

martinRenou: Yes it should also be added Python side

js/src/Map.js
martinRenou: Then maybe adding it would not be a good idea? If we need to remove it soon.

ipyleaflet/leaflet.py
betolink: Part of my experiments, going to revert it.

js/src/Map.js
betolink: I included it to experiment with my polar projections but had no effect on what I was trying to achieve at the moment. Maybe removing it altogether is the better option here.

js/src/leaflet.js
betolink: yup

js/src/Map.js
martinRenou: Agreed. But it could be added in a separate.

js/src/layers/WMSLayer.js
betolink: Should this use the `o=True` on the Python side? are you guys ok with `no_wrap: true` being the default?

martinRenou: > Should this use the o=True on the Python side?

Yes.

Actually this should be added to its base class `TileLayer`, and it should be false by default, so that it reflects the LeafletJS API: https://leafletjs.com/reference-1.6.0.html#gridlayer-nowrap

betolink: cool, I just noticed that it's already present on the base class, we just needed to use it on the js counterpart. All changes should be ready in a few...

js/src/Map.js
davidbrochart: I would be good to understand what it is for. The [web site](https://kartena.github.io/Proj4Leaflet) says: "note the use of the option continuousWorld for the tile layer, which tells Leaflet that the tiles should not be wrapped the same way they should for its default projection".

davidbrochart: But it's definitely not there anymore (https://github.com/Leaflet/Leaflet/pull/4058), so yes we should just remove it.

betolink:
I added eslint to the dev dependencies for javascript linting. Perhaps flake8 should also be part of the environment.yml file since the CI does linting on python.

js/src/Map.js
betolink: done!

ipyleaflet/leaflet.py
```diff
@@ -299,7 +300,7 @@ class WMSLayer(TileLayer):
     styles = Unicode().tag(sync=True, o=True)
     format = Unicode('image/jpeg').tag(sync=True, o=True)
     transparent = Bool(False).tag(sync=True, o=True)
-    crs = Enum(values=allowed_crs, default_value='EPSG3857').tag(sync=True)
+    crs = Dict(default_value=projections.EPSG3857).tag(sync=True, o=True)
```
martinRenou, line 303: You might need to remove `o=True` here as well.

js/src/leaflet.js
```diff
@@ -14,49 +16,49 @@ require('leaflet.awesome-markers');
...
           //L.DomEvent.off(tile, 'load', this._tileOnLoad); ... this doesnt work!
-          this._refreshTileUrl(tile,newsrc);
+          this._refreshTileUrl(tile, newsrc);
         }
       }
     }
 
     if (wasAnimated) {
-      setTimeout(function() { map._fadeAnimated = wasAnimated; }, 5000);
+      setTimeout(function () {this._map._fadeAnimated = wasAnimated;}, 5000);
```
martinRenou, line 49: I don't understand why this code was working before.

But I am also not sure it is still working now. Shouldn't this function be an arrow function?

martinRenou:
> I added eslint to the dev dependencies for javascript linting.

Would you mind adding a line in the `.travis.yml` file that runs the eslint check?

> Perhaps flake8 should also be part of the environment.yml file since the CI does linting on python.

The `environment.yml` file is for the binder link, it is not for a dev environment.

js/src/leaflet.js
betolink: Yeah I don't know why this is inside a timeout....

betolink:
> > I added eslint to the dev dependencies for javascript linting.
> 
> Would you mind adding a line in the `.travis.yml` file that runs the eslint check?
> 
> > Perhaps flake8 should also be part of the environment.yml file since the CI does linting on python.
> 
> The `environment.yml` file is for the binder link, it is not for a dev environment.

should this project have a environment-dev.yml? I think binder recommends putting their stuff inside a `binder` folder.  

I didn't add eslint to the CI because maybe fixing all that requires another PR.

martinRenou:
> should this project have a environment-dev.yml? I think binder recommends putting their stuff inside a binder folder.



> I didn't add eslint to the CI because maybe fixing all that requires another PR.

Agreed.

ipyleaflet/leaflet.py
betolink: I forgot this one, done!

js/src/leaflet.js
martinRenou: I don't remember the reason for the timeout. I remember this code was a bit weak but that was the only thing we could do.

Let's make it an arrow function anyway so the code still works, otherwise `this` is not defined properly.

martinRenou: See https://github.com/jupyter-widgets/ipyleaflet/pull/495 for reference

martinRenou: Actually I feel like this timeout thing never worked. Maybe this code is actually never triggered. Oo'

betolink: I'd say let's just leave it there until there is a PR for cleaning JS specific code.

betolink:
Thanks for the work on this widget, it's very useful!

scottyhq:
This is really neat, thanks @betolink. I'm I correct that this PR was just for the tile layer, but not geojson or image overlay features? 

It would be fantastic to support this for GeoJSON features https://ipyleaflet.readthedocs.io/en/latest/api_reference/geo_json.html --> https://github.com/kartena/Proj4Leaflet#lprojgeojson

And ImageOverlay
https://ipyleaflet.readthedocs.io/en/latest/api_reference/image_video_overlay.html# --> https://github.com/kartena/Proj4Leaflet#lprojimageoverlay

betolink:
Hi @scottyhq, geojson reprojection should work as long as you pass the proper CRS projection to the map constructor(either use the 2 polar projections included or define your own). So, from the example:

```python
from ipyleaflet import Map, GeoJSON, projections
import json
import os
import requests

if not os.path.exists('europe_110.geo.json'):
    url = 'https://github.com/jupyter-widgets/ipyleaflet/raw/master/examples/europe_110.geo.json'
    r = requests.get(url)
    with open('europe_110.geo.json', 'w') as f:
        f.write(r.content.decode("utf-8"))

with open('europe_110.geo.json', 'r') as f:
    data = json.load(f)

m = Map(center=(60, 0), zoom=2,  crs=projections.EPSG3413)
geo_json = GeoJSON(data=data, style = {'color': 'green', 'opacity':1, 'weight':1.9, 'dashArray':'9', 'fillOpacity':0.1})
m.add_layer(geo_json)

m

```

I haven't tested ImageOverlays yet, will do that next. 

Screenshot_2020-06-08 GeoJson_EU_on_hover - Jupyter Notebook

also, a virtual Hi! to the IceSat hack-week folks. 

Cheers!

Luis.

davidbrochart:
I guess GeoJSON works out of the box because the segments are small enough? I doubt we would see e.g. the great circle distance on an orthographic projection, but just a straight line between two points, right?
Similarly, I would think that for ImageOverlay only the bounds are handled, but no proper raster reprojection. For that, you could use rasterio like in https://github.com/jupyter-widgets/ipyleaflet/blob/master/examples/Numpy.ipynb.

scottyhq:
Thanks @betolink and @davidbrochart - I think I'll open up a separate issue to continue the discussion about overlaying data on custom projections. I imagine lots of people will be interested in this, so documenting what works and what doesn't will be of broad interest.

> also, a virtual Hi! to the IceSat hack-week folks. 

And a warm hello back! things are ramping up this week so definitely check out all the new content showing up here https://github.com/ICESAT-2HackWeek ;)

epifanio:
Hi @davidbrochart , @betolink ,

I am trying to understand how to define a custom projection but I am having a hard time to found the correct values to use.

In my use-case I need to add two Projected coordinate system for the Northern and southern hemisphere ( [EPSG:32761](https://epsg.io/32761) and [EPSG:32661](https://epsg.io/32661)  )

I tried with a reverse engineering approach to matching what we have in [pyleaflet/projections.EPSG3413](https://github.com/jupyter-widgets/ipyleaflet/blob/master/ipyleaflet/projections.py#L28)

```python
EPSG3413=dict(
    name='EPSG3413',       
    custom=True,        
    proj4def="""+proj=stere +lat_0=90 +lat_ts=70 +lon_0=-45 +k=1 +x_0=0 +y_0=0  +ellps=WGS84 +datum=WGS84 +units=m +no_defs""",        
    origin=[-4194304, 4194304],        
    resolutions=[16384.0, 8192.0, 4096.0, 2048.0, 1024.0, 512.0, 256.0 ],        
    bounds=[ [-4194304, -4194304], [4194304, 4194304] ] )
 ```

 with what I found on [epsg.io/2313](https://epsg.io/3413) 

but I failed in understanding how the values for `origin`, `resolution`, and `bounds` are computed.

Ideally,  it should be possible to code some sort of "mapping" to retrieve the information about a projection from `pyproj`, e.g. :

```python
from pyproj import CRS
crs = CRS.from_epsg(3413)
```

and use those information to generate an equivalent `Ipyleaflet.projection`.

As a use-case to help me understanding, do you have an idea on what should I use as projection in ipyleafelt to render the following WMS:

```
wms_url = "https://thredds.met.no/thredds/wms/sea/norkyst800m/24h/aggregate_be?SERVICE=WMS&REQUEST=GetCapabilities"
```

 in the WGS 84 / UPS North (N,E)  projection:
 
( [EPSG:32761](https://epsg.io/32761) and [EPSG:32661](https://epsg.io/32661)  )

betolink:
Hi @epifanio,

If I understood correctly, you want to use one of the layers in this WMS server projected in EPSG:32661. If this is the case then the parameters you're looking for would look like:

```python

from ipyleaflet import (
    Map,
    WMSLayer,
    DrawControl,
    projections
)

# so you can see the GeoJSON after you draw an object
def handle_draw(target, action, geo_json):
    print(action, geo_json)

EPSG_32661 = {
    'name': 'EPSG:32661',
    'custom': True, #This is important, it tells ipyleaflet that this projection is not on the predefined ones.
    'proj4def': '+proj=stere +lat_0=90 +lat_ts=90 +lon_0=0 +k=0.994 +x_0=2000000 +y_0=2000000 +datum=WGS84 +units=m +no_defs',
    'origin': [2000000, 424161],
    'bounds': [
        [2000000.0, 5405880.72],
        [2000000.0, 2000000.0]
    ],
    'resolutions': [
        16384.0,
        8192.0,
        4096.0,
        2048.0,
        1024.0,
        512.0,
        256.0
    ]
}

# Let's use the salinity layer
wms_layer = WMSLayer(
    url='https://thredds.met.no/thredds/wms/sea/norkyst800m/24h/aggregate_be',
    layers='salinity',
    format='image/png',
    transparent=True,
    min_zoom=1,
    attribution='The Norwegian Meteorological Institute as the source of data',
    crs=EPSG_32661 # I'm asking this WMS service to reproject the tile layer using EPSG:32661
)

# map center close to Norway
m = Map(center=(70, 15),
        zoom=0,
       max_zoom=6,
        layers=(wms_layer,),
        crs=EPSG_32661)

dc = DrawControl(marker={'shapeOptions': {'color': '#0000FF'}})
dc.on_draw(handle_draw)

m.add_control(dc)

m

```

The projection object is a simple dictionary,  getting the right values could be tricky as some projections are not well documented.
If you look at https://epsg.io/32661 you can find the projected bounds and a point of origin, these are the values you need.

* **origin**: the projected lon, lat that will be the center of the map
* **resolutions**: pixel resolution per zoom level  (this one is not in the EPSG def but let's use the same grid NASA uses for polar projections)
* **bounds**: projected max bounds  

Sometimes the values described in the epsg.io website are not always correct. i.e. in this projection the bounds are a little off. The bounds are used by leaflet to calculate which tiles are going to be requested to the WMS server. They don't need to align perfectly but... if you get out of bounds you may requests tiles that WMS cannot produce for you and hence a lot of 404s will show up.

This is the layer I see on ipyleaflet:




I hope this is what you were looking for.

epifanio:
Hi @betolink , thank you so much for the help!

I had two main problems which you helped me to fix:

1) I was setting the projection using the value you indicated, for the `ipyleaflet.WMSlayer` .. but not for the `ipyleaflet.Map`

2) I was trying to code a method that can return those values, given the EPSG code on the WMS projection - and while doing so I was using the wrong `pyproj` method to perform the conversion from geographic to projected (for the bounds)



```python
from pyproj import CRS, Transformer
epsg_code = 32661

# transformer = Transformer.from_crs("epsg:4326", f"epsg:{epsg_code}")  <-- wrong
crs = CRS.from_epsg(epsg_code)
transformer = Transformer.from_crs(crs.geodetic_crs, crs)    #  <-- correct (or at least seems so)
```

Note: I noticed the bounds you suggest are the same one used in  https://spatialreference.org which also agrees with the one returned from the `pyproj` transformation.


At this point, I had to find the projected coordinates for the origin ( I wonder if I am totally mistaken it .. is it intended to be the center of the projected coordinates bounds?)  -- but my values are off from what I found on the epsg.io and from what you used (all 3 are very different  to each other)

I ended up with this code:


```python
from pyproj import CRS,Transformer

def get_center(ll, ur):
    c_lat = ((max(ll[0], ur[0]) - min(ll[0], ur[0]))/2) + min(ll[0], ur[0])
    c_lon = ((max(ll[1], ur[1]) - min(ll[1], ur[1]))/2) + min(ll[1], ur[1])
    return [c_lon, c_lat]

def get_projection(epsg_code):
    crs = CRS.from_epsg(epsg_code)
    proj = Transformer.from_crs(crs.geodetic_crs, crs)
    ll_lon, ll_lat, ur_lon, ur_lat = crs.area_of_use.bounds
    ll = proj.transform(ll_lat, ll_lon)
    ur = proj.transform(ur_lat, ur_lon)
    origin = get_center(ll, ur)
    ipyl_projection = {
        "name": f"EPSG:{epsg_code}",
        "custom": True,
        "proj4def": crs.to_proj4(),
        "origin": origin,
        "bounds": [[ll[1], ll[0]], [ur[1], ur[0]]],
        "resolutions": [16384.0, 8192.0, 4096.0, 2048.0, 1024.0, 512.0, 256.0],
    }
    return ipyl_projection
```

which will return a dictionary that could be used as an `Ipyleaflet.projection`:

```python
get_projection(32661)
```
```python
{'name': 'EPSG:32661',
 'custom': True,
 'proj4def': '+proj=stere +lat_0=90 +lon_0=0 +k=0.994 +x_0=2000000 +y_0=2000000 +datum=WGS84 +units=m +no_defs +type=crs',
 'origin': [1999999.9999999998, 3702940.358685655],
 'bounds': [[1999999.9999999995, 5405880.71737131], [2000000.0, 2000000.0]],
 'resolutions': [16384.0, 8192.0, 4096.0, 2048.0, 1024.0, 512.0, 256.0]}
 ```

I can see two main issues remaining, I am not sure about the value used as 'origin' , the 'resolution' is hardcoded.

Testing an overlay with the Geojson bounds, it seems to match. 

image


The main reason for trying to automate the building of the projection, is to be able to overlay  different WMS, and let the user to switch between the available projections (when a set of common projection is found in their GetCapabilities)

I am doing this in a little wms viewer based on ipyleaflet, to which I need to add a widget to switch between projections

 e.g.: early draft for 3 WMS, using only 4326 :  

https://bokeh.epinux.com/GISPY?url=http://nbswms.met.no/thredds/wms_ql/NBS/S1A/2021/05/18/EW/S1A_EW_GRDM_1SDH_20210518T070428_20210518T070534_037939_047A42_65CD.nc?SERVICE=WMS&REQUEST=GetCapabilities&url=https://thredds.met.no/thredds/wms/cmems/si-tac/cmems_obs-si_arc_phy-siconc_nrt_L4-auto_P1D_aggregated?service=WMS&version=1.3.0&request=GetCapabilities&url=https://thredds.met.no/thredds/wms/sea/norkyst800m/24h/aggregate_be

Issue #591, opened 2020-05-13
To make the float zoom levels useful for Vector Tile Layer


omanges:
Even though the zoom level trait has been made Float in the Vector Tile Layer support feature (https://github.com/jupyter-widgets/ipyleaflet/pull/544), the zoom only changes on a integer value, for example, map actually changes only on 1.0, 2.0, 3.0 ..., rather than changing on any floating zoom, as floating zoom levels are possible for Vector Tile.

davidbrochart:
Looks like we need support for https://leafletjs.com/reference-1.6.0.html#map-zoomdelta

martinRenou:
It should be straightforward to add expose this `zoom_delta` option to Python

omanges:
@martinRenou  assign this to me, I can try this.

omanges:
I think we should also implement the https://leafletjs.com/reference-1.6.0.html#map-zoomsnap, zoomsnap

omanges:
Work for me now 
![image](https://user-images.githubusercontent.com/30625612/83386930-7b637d00-a409-11ea-89b0-e34596da8787.png)

zxdawn:
Is it possible to make it also work for the scroll wheel?

PR #608, opened 2020-05-31
Changes to enable float zoom levels

omanges:
These changes fix the issue raised in https://github.com/jupyter-widgets/ipyleaflet/issues/591


martinRenou:
LGTM! I can squash and merge.

martinRenou:
The CI failure seems unrelated, but I am not sure

martinRenou:
Yes it's unrelated. It might be a new flake8 rule. Let's fix it in another PR.

davidbrochart:
`zoom_delta` and `zoom_snap` and not used on the JS side?

martinRenou:
It is :) It will be auto-magically handled by ipyleaflet, being an option (`0=True`).

davidbrochart:
Did you check that it works fine? I tried on a vector tile layer and it doesn't seem to work.

omanges:
Yes, it is working for me.
[image: image.png]

martinRenou:
> Did you check that it works fine? I tried on a vector tile layer and it doesn't seem to work.

Actually you're right. Handling options changes is not implemented on the `Map`. This should be fixed.

davidbrochart:
@omanges it looks like it works for everything but a vector tile layer.

davidbrochart:
My bad, I didn't pass the options on the `Map` but on the layer!
Everything's good, sorry!

omanges:
Great, It works for you as well!!!

PR #611, opened 2020-06-01
Fix option update

### LINKED ISSUE #610: Handle options changes
Only the Controls properly handles options changes.

We should have something like:
```javascript
model_events() {
  let key;
  let options = this.model.get('options');
  for (var i = 0; i < o.length; i++) {
    key = options[i];
    this.listenTo(this.model, 'change:' + key, () => {
    }, this);
  }
}
```

On the Layer base class and on the Map.

davidbrochart:
Fixes #610

### DIFF
--- a/js/src/layers/Path.js
+++ b/js/src/layers/Path.js
@@ -27,6 +27,19 @@ export class LeafletPathModel extends vectorlayer.LeafletVectorLayerModel {
 export class LeafletPathView extends vectorlayer.LeafletVectorLayerView {
   model_events() {
     super.model_events();
+    var key;
+    var o = this.model.get('options');
+    for (var i = 0; i < o.length; i++) {
+      key = o[i];
+      this.listenTo(
+        this.model,
+        'change:' + key,
+        function() {
+          this.obj.setStyle(this.get_options());
+        },
+        this
+      );
+    }
     this.obj.setStyle(this.get_options());
   }
 }
--- a/js/src/layers/Path.js
+++ b/js/src/layers/Path.js
@@ -27,19 +27,6 @@ export class LeafletPathModel extends vectorlayer.LeafletVectorLayerModel {
 export class LeafletPathView extends vectorlayer.LeafletVectorLayerView {
   model_events() {
     super.model_events();
-    var key;
-    var o = this.model.get('options');
-    for (var i = 0; i < o.length; i++) {
-      key = o[i];
-      this.listenTo(
-        this.model,
-        'change:' + key,
-        function() {
-          this.obj.setStyle(this.get_options());
-        },
-        this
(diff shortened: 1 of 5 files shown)


martinRenou:
Does it work correctly?

There is now some code duplication, `layers/Velocity.js`, `layers/Heatmap.js` and `layers/Path.js` seem to implement it already so we might want to remove it there.

davidbrochart:
> Does it work correctly?

Yes I checked on the Map and some layers and it seems to work fine.

> There is now some code duplication, `layers/Velocity.js`, `layers/Heatmap.js` and `layers/Path.js` seem to implement it already so we might want to remove it there.

Good point.

js/src/layers/Path.js
```diff
@@ -27,19 +27,6 @@ export class LeafletPathModel extends vectorlayer.LeafletVectorLayerModel {
...
     super.model_events();
-    var key;
-    var o = this.model.get('options');
-    for (var i = 0; i < o.length; i++) {
-      key = o[i];
-      this.listenTo(
-        this.model,
-        'change:' + key,
-        function() {
-          this.obj.setStyle(this.get_options());
```
martinRenou, line 38: Did you check that this was still working? It's not exactly the same code.

davidbrochart: Good point, it doesn't work. It works if I keep `Path.js` as it were, but the `LeafletLayerView` still calls `L.setOptions` on the `LeafletPathView`. It doesn't seem to hurt, but is it okay?

martinRenou: Let's keep it as it was then.. It is a bit annoying. The API of LeafletJS is too heterogeneous.

davidbrochart: Do you mean we should close this PR?

martinRenou: No I mean let's keep `Path.js` as it was with the `setStyle`, but keep your other changes.

davidbrochart: :+1:

js/src/layers/Path.js
```diff
@@ -27,6 +27,19 @@ export class LeafletPathModel extends vectorlayer.LeafletVectorLayerModel {
...
+      this.listenTo(
+        this.model,
+        'change:' + key,
+        function() {
+          this.obj.setStyle(this.get_options());
+        },
+        this
+      );
+    }
     this.obj.setStyle(this.get_options());
```
martinRenou: This line is unexpected :open_mouth: It should not be needed

martinRenou: Let's remove it later

davidbrochart: :+1:

PR #616, opened 2020-06-12
Add pixel_bounds to Map

davidbrochart:
Similar to the `bounds` read-only attribute, but in projected pixel coordinates instead of lat/lon coordinates (see https://leafletjs.com/reference-1.0.3.html#map-getpixelbounds).
While `bounds` is `((south, west), (north, east))`, `pixel_bounds` is `((left, top), (right, bottom))` to respect the Leaflet convention.

### DIFF
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -1032,6 +1032,11 @@ class Map(DOMWidget, InteractMixin):
     east = Float(def_loc[1], read_only=True).tag(sync=True)
     west = Float(def_loc[1], read_only=True).tag(sync=True)
 
+    bottom = Int(0, read_only=True).tag(sync=True)
+    top = Int(9007199254740991, read_only=True).tag(sync=True)
+    right = Int(0, read_only=True).tag(sync=True)
+    left = Int(9007199254740991, read_only=True).tag(sync=True)
+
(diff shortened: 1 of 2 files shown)


martinRenou:
Is it the size in pixels of the map?

davidbrochart:
No, it's bounds in pixels.

martinRenou:
In screen pixels? Is it the size of the map?

davidbrochart:
Yes, screen pixels. You can get the size of the map from it.

martinRenou:
But that does not work when having multiple views with different sizes

davidbrochart:
As for the bounds, we take the biggest view.

martinRenou:
I see :) ok

ipyleaflet/leaflet.py
```diff
@@ -1032,6 +1032,11 @@ def _default_options(self):
     east = Float(def_loc[1], read_only=True).tag(sync=True)
     west = Float(def_loc[1], read_only=True).tag(sync=True)
 
+    bottom = Int(0, read_only=True).tag(sync=True)
+    top = Int(9007199254740991, read_only=True).tag(sync=True)
```
martinRenou, line 1036: Why this default value?

davidbrochart: It's the biggest `Number`. We need this initial value in order to play well with the `min`.

Issue #618, opened 2020-06-14
Removing layer also removes basemap


davidbrochart:
I've not investigated yet, but it seems that removing a layer leaves a blank map (on master):

```python
from ipyleaflet import Map, basemaps, basemap_to_tiles

m = Map(center=(52, 10), zoom=8, basemap=basemaps.CartoDB.DarkMatter)
m

strava_all = basemap_to_tiles(basemaps.Strava.All)
m.add_layer(strava_all)

m.remove_layer(strava_all)
```

sackh:
@davidbrochart  and @martinRenou - I have raised PR #619 please review.

PR #619, opened 2020-06-15
Fix for issue #618 - Removing layer also removes basemap

sackh:
Hi @martinRenou 
I have fixed this issue, I think this has happened when we fixed flake 8 errors.
I have renamed the variables now, it is working fine.


martinRenou:
Indeed thanks!

davidbrochart:
Awesome @sackh, I checked that it works fine.

Issue #612, opened 2020-06-09
How to Add Data Layers on Custom Projections


scottyhq:
https://github.com/jupyter-widgets/ipyleaflet/pull/598 added the ability to display maps with custom projections, which is highly useful for the polar regions! 

It's a bit unclear currently what is supported for adding data layers onto maps with custom projections. According to the [Proj4Leaflet readme](https://github.com/kartena/Proj4Leaflet) it seems straightforward to support GeoJSON and Image Overlays, but as far as I can tell this isn't yet supported (ipyleaflet 0.13.0).

Here is a full gist (https://gist.github.com/scottyhq/39b39917e851a2fc691fa6e6d9716794) with a few self-contained examples for adding a coastline layer or simple bounding box for the south pole (projections.EPSG3031) 

Default projection
<img width="824" alt="Screenshot 2020-06-09 10 28 35" src="https://user-images.githubusercontent.com/3924836/84181587-fc3ffb80-aa3d-11ea-8da5-b06222a298d1.png">

SPS projection
<img width="627" alt="Screenshot 2020-06-09 10 28 24" src="https://user-images.githubusercontent.com/3924836/84181528-e8949500-aa3d-11ea-8311-620523705feb.png">
And for the north polar projection (projections.EPSG3413) here is an example from @betolink with screenshot showing offset polygons

```python
from ipyleaflet import Map, GeoJSON, projections
import json
import os
import requests

if not os.path.exists('europe_110.geo.json'):
    url = 'https://github.com/jupyter-widgets/ipyleaflet/raw/master/examples/europe_110.geo.json'
    r = requests.get(url)
    with open('europe_110.geo.json', 'w') as f:
        f.write(r.content.decode("utf-8"))

with open('europe_110.geo.json', 'r') as f:
    data = json.load(f)

m = Map(center=(60, 0), zoom=2,  crs=projections.EPSG3413)
geo_json = GeoJSON(data=data, style = {'color': 'green', 'opacity':1, 'weight':1.9, 'dashArray':'9', 'fillOpacity':0.1})
m.add_layer(geo_json)

m
```
<img width="890" alt="Screenshot 2020-06-09 10 38 11" src="https://user-images.githubusercontent.com/3924836/84181394-b5ea9c80-aa3d-11ea-9ed3-42672f01855b.png">

cc @davidbrochart @asteiker

davidbrochart:
[...]

PR #620, opened 2020-06-15
Fix EPSG:3413 and EPSG:3031

davidbrochart:
This is applying [this change](https://github.com/jupyter-widgets/ipyleaflet/issues/612#issuecomment-641761444).
@betolink I also changed the comment to `origin and bounds for 250m resolution` instead of 500m, is this correct?
Should we do the same for `EPSG:3031`?

### DIFF
--- a/ipyleaflet/projections.py
+++ b/ipyleaflet/projections.py
@@ -51,9 +51,10 @@ projections = Bunch(
         custom=True,
         proj4def="""+proj=stere +lat_0=-90 +lat_ts=-71 +lon_0=0 +k=1 +x_0=0 +y_0=0
                  +ellps=WGS84 +datum=WGS84 +units=m +no_defs""",
-        # origin and bounds for 500m resolution
-        origin=[-4194304 / 2, 4194304 / 2],
+        # origin and bounds for 250m resolution
+        origin=[-4194304, 4194304],
         resolutions=[
+            16384.0,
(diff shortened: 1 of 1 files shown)


betolink:
Thanks for this @davidbrochart! yeah the polar stereo south needs the same changes.

davidbrochart:
Thanks!
Although I don't really understand the comment `origin and bounds for 250m resolution`. Shouldn't it be the smallest resolution, in this case `256m`?

betolink:
oh that comment, I'm talking about the basemap there:  NASA's blue marble next gen native resolution is 500m using polar projections but that's not necessary the case for other layers or other providers. We need to include all the resolutions just in case a user decides to use a different base map. So 500m ≈ 512.0 m/per pixel https://wiki.earthdata.nasa.gov/display/GIBS/GIBS+API+for+Developers

betolink:
Ultimately if a user is using a custom projection is a good idea to define their own bounds/resolutions/origin just because 3413 and 3013 are not even the only polar projections out there.

davidbrochart:
I see but the basemap (e.g. `basemaps.NASAGIBS.BlueMarble3413`) is provided separately in the Map, so maybe we should just get rid of this comment?

betolink:
yeah get rid of the comment! I should work on expanding the example notebook!

PR #621, opened 2020-06-15
Implement __geo_interface__ for relevant classes

vpipkt:
This PR implements the [`__geo_interface__` spec][spec] for `GeoJSON`, `Choropleth`, and `GeoData`. For `GeoData` we fall back to `geopandas` implementation on the `GeoDataFrame`.

Note that as is the convention with GeoDataFrame a` {'type': 'FeatureCollection', ...}` can be returned. This is not explicitly stated in the spec but the comments lobby for it and ask the geopandas contributors to expand on their approach.

[spec]: https://gist.github.com/sgillies/2217756

This is my first contribution to the project so I appreciate your attention and  patience!

### DIFF
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -545,6 +545,16 @@ class GeoJSON(FeatureGroup):
 
         return self.data
 
+    @property
+    def __geo_interface__(self):
+        """
+        Return a dict whose structure aligns to the GeoJSON format
+        For more information about the ``__geo_interface__``, see
+        https://gist.github.com/sgillies/2217756
+        """
+
+        return self.data
+
     def _apply_style(self, feature, style_callback):
         if 'properties' not in feature:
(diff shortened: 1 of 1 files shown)


ipyleaflet/leaflet.py
```diff
@@ -545,6 +545,16 @@ def _get_data(self):
 
         return self.data
 
+    @property
+    def __geo_interface__(self):
+        """
```
martinRenou, line 550: Thanks for making a docstring! We need more of those.

davidbrochart:
Awesome @vpipkt !
This is good to merge after fixing the linting issue.

vpipkt:
@martinRenou Yes I would say it would be great for Marker and Polygon. Should I add to this PR or make a separate one?

martinRenou:
Another PR is fine :) Thanks again!

Issue #585, opened 2020-05-11
Add layer attribute to the SearchControl


martinRenou:
This would permit searching through *e.g.* a GeoJSON layer

sackh:
@martinRenou - please close this as this is already done in #622.

PR #622, opened 2020-06-15
Layer attribute search control

sackh:
Hi @martinRenou 
I have created a new PR for issue #585  as the previous one got closed by mistake.
Please review this.


js/src/controls/SearchControl.js
```diff
@@ -19,22 +19,43 @@ export class LeafletSearchControlModel extends control.LeafletControlModel {
...
-      options.marker = view.obj;
-      this.obj = L.control.search(options);
-    });
+    if (this.model.get('layer') !== null) {
+        return this.create_child_view(this.model.get('layer')).then((layer_view) => {
+            let options = this.get_options();
+            options.layer = layer_view.obj;
+            options.marker = false;
+            this.obj = L.control.search(options);
+        });
```
martinRenou, line 42: It looks like your implementation makes it impossible to set the `marker` and the `layer` attributes at the same time.

What about something like this for the `create_obj` method:
```javascript
const layer = this.model.get('layer');
const marker = this.model.get('marker');

const layer_promise = layer !== null ? this.create_child_view(layer) : Promise.resolve(null);
const marker_promise = marker !== null ? this.create_child_view(marker) : Promise.resolve(null);

return Promise.all([layer_promise, marker_promise]).then((layer_view, marker_view) => {
    const options = this.get_options();
    options.layer = layer_view !== null ? layer_view.obj : false;
    options.marker = marker_view !== null ? marker_view.obj : false;
    this.obj = L.control.search(options);
});
```

I did not try it locally so you might need to change one or two things.

js/src/controls/SearchControl.js
```diff
@@ -19,22 +19,43 @@ export class LeafletSearchControlModel extends control.LeafletControlModel {
...
+            options.marker = view.obj;
+            this.obj = L.control.search(options);
+        });
+    }
+  }
+
+  leaflet_events() {
+    if (this.model.get('layer') !== null) {
+        this.obj.on('search:locationfound', function(e) {
+            e.layer.setStyle({fillColor: '#3f0', color: '#0f0'});
```
martinRenou, line 55: Should we remove this style once the search control searches for something else?

sackh: Currently, once you clicked on the searched feature style gets removed automatically. I will have to check if there is any other event to clear the style.

js/src/controls/SearchControl.js
sackh: I will try this.

martinRenou: You could indeed do that if the user did not define a Marker? But if the user defined a Marker I guess they will expect it to be used.

sackh: Sure, I will try the changes you suggested.

sackh: @martinRenou - updated the code as suggested by you.

martinRenou:
We might want to make the "found" style an attribute, but let's do it later :) Thanks a lot for your work!

Issue #633, opened 2020-06-23
KeyError: 'popup_anchor' for Icon


sackh:
version: master branch 

Code to reproduce Error:
```python
from ipyleaflet import Marker, Icon, Map

center = (52.204793, 360.121558)

m = Map(center=center, zoom=10)
icon = Icon(icon_url='https://leafletjs.com/examples/custom-icons/leaf-green.png', icon_size=[38, 95], icon_anchor=[22,94])
mark = Marker(location=center, icon=icon, rotation_angle=90, rotation_origin='22px 94px')
m.add_layer(mark);

m
```
Error:
```
KeyError                                  Traceback (most recent call last)
~/miniconda3/envs/ipyleafmy2/lib/python3.7/site-packages/traitlets/traitlets.py in get(self, obj, cls)
    527         try:
--> 528             value = obj._trait_values[self.name]
    529         except KeyError:

KeyError: 'popup_anchor'

During handling of the above exception, another exception occurred:

TraitError                                Traceback (most recent call last)
<ipython-input-1-fc0c993d181a> in <module>
      4 
      5 m = Map(center=center, zoom=10)
----> 6 icon = Icon(icon_url='https://leafletjs.com/examples/custom-icons/leaf-green.png', icon_size=[38, 95], icon_anchor=[22,94])
      7 mark = Marker(location=center, icon=icon, rotation_angle=90, rotation_origin='22px 94px')
      8 m.add_layer(mark);

~/pers/hubs/forks/ipyleaflet/ipyleaflet/leaflet.py in __init__(self, **kwargs)
    119 
    120     def __init__(self, **kwargs):
--> 121         super(Layer, self).__init__(**kwargs)
    122         self.on_msg(self._handle_mouse_events)
    123 

~/miniconda3/envs/ipyleafmy2/lib/python3.7/site-packages/ipywidgets/widgets/widget.py in __init__(self, **kwargs)
    413 
    414         Widget._call_widget_constructed(self)
--> 415         self.open()
    416 
    417     def __del__(self):

~/miniconda3/envs/ipyleafmy2/lib/python3.7/site-packages/ipywidgets/widgets/widget.py in open(self)
    426         """Open a comm to the frontend if one isn't already open."""
    427         if self.comm is None:
--> 428             state, buffer_paths, buffers = _remove_buffers(self.get_state())
    429 
    430             args = dict(target_name='jupyter.widget',

[...]

PR #634, opened 2020-06-23
Fix for KeyError: 'popup_anchor' and other traits of Icon

sackh:
Fix for issue #633

### DIFF
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -260,11 +260,11 @@ class Icon(UILayer):
 
     icon_url = Unicode('').tag(sync=True, o=True)
     shadow_url = Unicode(None, allow_none=True).tag(sync=True, o=True)
-    icon_size = List(None, allow_none=True, minlen=2, maxlen=2).tag(sync=True, o=True)
-    shadow_size = List(None, allow_none=True, minlen=2, maxlen=2).tag(sync=True, o=True)
-    icon_anchor = List(None, allow_none=True, minlen=2, maxlen=2).tag(sync=True, o=True)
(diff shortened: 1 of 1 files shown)


martinRenou:
This is weird. It should not behave that way IMO. The `default_value` is set to `None`, and we allow the `None` value. Also, `null` is the default in LeafletJS, so I think we should respect that.

It might be an issue with traitlets itself.

I agree this issue should be fixed, I am just unsure with this fix.

sackh:
One more observation is that when we pass parameters `minlen=2, maxlen=2` it fails even though default_value is set to `None`, and we allow the `None`,  but if we remove the `minlen=2, maxlen=2` then it works. It looks like an issue with traitlets when we pass params `minlen=2, maxlen=2`

martinRenou:
I guess it's unlikely that we can fix Traitlets and get a release soon.

What we could do is implement our own traitlets validator, checking the length ourselves. What do you think?

Thanks for spotting this issue BTW, we should fix this issue before making a release.

sackh:
Yes, Traitlets validator is a good idea. shall we use `validate` decorator from Traitlets itself or any other approach?

martinRenou:
The `validate` decorator is fine I guess :)

sackh:
Ok, I will make these changes and raise the PR.

SylvainCorlay:
Thanks for finding a workaround!

martinRenou:
Why did you discard the comments @SylvainCorlay ?

martinRenou:
I guess you missed them. I opened a PR to fix this. It appears to be a Traitlets bug though.. Maybe we should try to fix it upstream.

PR #636, opened 2020-06-28
Fix icon attrs default values

martinRenou:
Workaround for the discussion in #634

### DIFF
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -260,11 +260,24 @@ class Icon(UILayer):
 
     icon_url = Unicode('').tag(sync=True, o=True)
     shadow_url = Unicode(None, allow_none=True).tag(sync=True, o=True)
-    icon_size = List(None, default_value=[0, 0], allow_none=True, minlen=2, maxlen=2).tag(sync=True, o=True)
-    shadow_size = List(None, default_value=[0, 0], allow_none=True, minlen=2, maxlen=2).tag(sync=True, o=True)
-    icon_anchor = List(None, default_value=[0, 0], allow_none=True, minlen=2, maxlen=2).tag(sync=True, o=True)
-    shadow_anchor = List(None, default_value=[0, 0], allow_none=True, minlen=2, maxlen=2).tag(sync=True, o=True)
-    popup_anchor = List(None, default_value=[0, 0], allow_none=True, minlen=2, maxlen=2).tag(sync=True, o=True)
+    icon_size = List(default_value=None, allow_none=True).tag(sync=True, o=True)
+    shadow_size = List(default_value=None, allow_none=True).tag(sync=True, o=True)
+    icon_anchor = List(default_value=None, allow_none=True).tag(sync=True, o=True)
+    shadow_anchor = List(default_value=None, allow_none=True).tag(sync=True, o=True)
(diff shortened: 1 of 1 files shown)


ipyleaflet/leaflet.py
```diff
@@ -260,11 +260,24 @@ class Icon(UILayer):
...
-    icon_size = List(None, default_value=[0, 0], allow_none=True, minlen=2, maxlen=2).tag(sync=True, o=True)
-    shadow_size = List(None, default_value=[0, 0], allow_none=True, minlen=2, maxlen=2).tag(sync=True, o=True)
-    icon_anchor = List(None, default_value=[0, 0], allow_none=True, minlen=2, maxlen=2).tag(sync=True, o=True)
-    shadow_anchor = List(None, default_value=[0, 0], allow_none=True, minlen=2, maxlen=2).tag(sync=True, o=True)
-    popup_anchor = List(None, default_value=[0, 0], allow_none=True, minlen=2, maxlen=2).tag(sync=True, o=True)
+    icon_size = List(default_value=None, allow_none=True).tag(sync=True, o=True)
+    shadow_size = List(default_value=None, allow_none=True).tag(sync=True, o=True)
+    icon_anchor = List(default_value=None, allow_none=True).tag(sync=True, o=True)
+    shadow_anchor = List(default_value=None, allow_none=True).tag(sync=True, o=True)
+    popup_anchor = List([0, 0], allow_none=True).tag(sync=True, o=True)
```
davidbrochart, line 267: Why is the default value not `None`?

ipyleaflet/leaflet.py
```diff
@@ -260,11 +260,24 @@ class Icon(UILayer):
...
+
+    @validate('icon_size', 'shadow_size', 'icon_anchor', 'shadow_anchor', 'popup_anchor')
+    def _validate_attr(self, proposal):
+        value = proposal['value']
+
+        # Workaround Traitlets which does not respect the None default value
+        if len(value) == 0:
```
davidbrochart, line 274: Won't it fail if the value is `None`?

ipyleaflet/leaflet.py
martinRenou: I am trying to respect the LeafletJS default values: https://leafletjs.com/reference-1.6.0.html#icon-popupanchor

ipyleaflet/leaflet.py
martinRenou: No, Traitlets seems to have set the value to an empty list `[]` in the meantime.

ipyleaflet/leaflet.py
davidbrochart: :+1:

ipyleaflet/leaflet.py
davidbrochart: :+1:

martinRenou: Maybe this line should be `if value is None or len(value) == 0:` to handle the case when Traitlets will be fixed, if that happens

davidbrochart: Yes, that would be safer.

ipyleaflet/leaflet.py
martinRenou: This makes me think I forgot to update them JS side #637

PR #637, opened 2020-06-28
Fix Icon layer default attrs JS side

martinRenou:


martinRenou:
@davidbrochart Please have a look and merge it if you think it's good :)

PR #640, opened 2020-07-01
Added found_style attribute for SearchControl

sackh:

Hi @martinRenou,

Added `found_style` attribute for SearchControl, now the user can specify his own style for the searched feature.
Please review.


js/src/controls/SearchControl.js
```diff
@@ -50,8 +51,13 @@ export class LeafletSearchControlView extends control.LeafletControlView {
 
   leaflet_events() {
     if (this.model.get('layer') !== null) {
+        var found_style = this.model.get('found_style');
         this.obj.on('search:locationfound', function(e) {
-            e.layer.setStyle({fillColor: '#3f0', color: '#0f0'});
+            if (Object.keys(found_style).length != 0) {
+                e.layer.setStyle(found_style);
+            } else {
+               e.layer.setStyle({fillColor: '#3f0', color: '#0f0'});
```
martinRenou, line 59: It might be nicer to have this as the default value for `found_style` Python side. So that if people wants the found feature to not be styled (and only use the marker) then they can set `found_style` to empty dict and there is no extra styling. What do you think?

sackh: yes, this makes sense.  I will do the required change.

js/src/controls/SearchControl.js
```diff
@@ -18,6 +18,7 @@ export class LeafletSearchControlModel extends control.LeafletControlModel {
       auto_collapse: false,
       zoom:10,
       animate_location:false,
+      found_style: {},
```
martinRenou, line 21: The default value JavaScript side does not match the one Python side (this makes a difference when using embedded widgets)

sackh: Done.

martinRenou:
Thanks a lot!

PR #662, opened 2020-07-29
Build docs in the CI

martinRenou:


martinRenou:
The build is failing as intended. It should be fixed by https://github.com/jupyter-widgets/ipyleaflet/pull/661

PR #665, opened 2020-07-30
Update image_video_overlay.rst

konanast:
Improve documentation for image_video_overlay

### DIFF
--- a/docs/source/api_reference/image_video_overlay.rst
+++ b/docs/source/api_reference/image_video_overlay.rst
@@ -44,6 +44,6 @@ Attributes
 ===========    ========================   ===
 Attribute      Default Value              Doc
 ===========    ========================   ===
-url            ""                         An http url to the footage or a relative path to a local file (image/video). Note that absolute local paths are not supported. 
+url            ""                         An http url to the footage or a relative path to a local file (image/video, e.g. ‘image.png’ or ‘files/working_dir/image.png’ in JupyterLab). Note that absolute local paths are not supported. 
 bounds         ((0.0, 0.0), (0.0, 0.0))   SW and NE corners of the image
 ===========    ========================   ===


docs/source/api_reference/image_video_overlay.rst
```diff
@@ -44,6 +44,6 @@ Attributes
 ===========    ========================   ===
 Attribute      Default Value              Doc
 ===========    ========================   ===
-url            ""                         An http url to the footage or a relative path to a local file (image/video). Note that absolute local paths are not supported. 
+url            ""                         An http url to the footage or a relative path to a local file (image/video, e.g. ‘image.png’ or ‘files/working_dir/image.png’ in JupyterLab). Note that absolute local paths are not supported.
```
davidbrochart, line 47: What is different in JupyterLab, the `files/` prefix?

martinRenou: Yes, otherwise it does not work.

martinRenou:
The CI error is very odd. Definitely not related to the changes here though.

docs/source/api_reference/image_video_overlay.rst
davidbrochart: Maybe we should make it clearer, like: "Note that in JupyterLab the path must be prefixed with `files/`, e.g. `files/image.png`".

konanast: Also important is the working_dir so you have to add "files/my_working_dir/", this is a bit strange and it is not mentioned anywhere.

martinRenou: It is mentioned there in the documentation: https://ipyleaflet.readthedocs.io/en/latest/api_reference/local_tile_layer.html#example

martinRenou: In Jupyter Notebook the path is relative to the Notebook. In JupyterLab it's relative to the server.

davidbrochart: > Also important is the current working folder name so you have to add "files/my_working_dir/", this is a bit strange and it is not mentioned anywhere.

Yes but this is not specific to JupyterLab, so we should make it consistent with the notebook.

davidbrochart: > In Jupyter Notebook the path is relative to the Notebook. In JupyterLab it's relative to the server.

Thanks, then this should also be mentioned.

Issue #669, opened 2020-08-02
Add support for  KML, GPX, TopoJSON, WKT formats


sackh:
library: https://github.com/mapbox/leaflet-omnivore
@martinRenou  - Let me know your thoughts, I can open PR for this feature.

davidbrochart:
Martin won't be available for the next two weeks. I can review a PR if you feel like opening one. I think it can be interesting to have this feature.

sackh:
@martinRenou - for `KML` and `GPX`,  files can be read into `Geopandas` Dataframe, shall I add separate layers for these or just adding examples in notebooks with `GeoData` layer for these file formats is sufficient?

giswqs:
Be cautious about adding `Geopandas` as a dependency of `ipyleaflet`. `Geopandas` has `GDAL` dependency, which is challenging for Windows users to install.

davidbrochart:
It might be worth looking at [kml2geojson](https://github.com/mrcagney/kml2geojson).

davidbrochart:
Or warn the user that if he needs this feature he needs to install geopandas, but without making it a dependency of ipyleaflet.

sackh:
I was suggesting that if the user reads files in GeoPandas DataFrame and uses our `GeoData` layer to add it over the map is it sufficient? or do we want to explicitly support KML and Gpx layers?

davidbrochart:
I guess it is fine to let users read files with GeoPandas DataFrames and use the GeoData layer.

PR #673, opened 2020-08-08
Added WKT layer feature to add WKT data on map

sackh:


sackh:
@davidbrochart  - please review this. I have added WKT support as mentioned in #669

davidbrochart:
Thanks. I'm wondering if we should rather use [shapely](https://shapely.readthedocs.io/en/latest/manual.html) to do these conversions on the Python side, and always use a `GeoJSON` layer.

sackh:
Yes, shapely would be good for `wkt` to `GeoJSON` conversion, I will check whether it supports other formats as well.

sackh:
closing this I have opened new PR #679 using shapely to convert WKT to GeoJSON on python side.

Issue #687, opened 2020-08-26
Have no default for zoom param in SearchControl?


deeplook:
At the moment when using `SearchControl` one can pass a `zoom` parameter to indicate the zoom level after jumping to the location. If not provided the default value `10` will be used. It would be more useful, I find, to have no specific default value, but leave the zoom level unchanged when not passing this parameter. Any thoughts?

martinRenou:
I agree. In the SeachControl plugin the default zoom value is `null` (see the [docs](https://github.com/stefanocudini/leaflet-search#options)), I suspect passing `null` leaves the zoom level unchanged. We might want to have a `None` default value Python side and `null` JavaScript side to keep this behavior.

sackh:
This makes sense to me as well, I will do this change and raise PR.

deeplook:
Cool, thanks!!

PR #688, opened 2020-08-26
Made default value of zoom trait for SearchControl None, so it will not change by default

sackh:
@martinRenou - I have tested this, please review.
fix for #687


martinRenou:
Great! Thanks!

Issue #702, opened 2020-09-23
Access the data of the feature found by a GeoJSON-based SearchControl?


AnotherKamila:
Apologies if this is documented somewhere, but I have been unable to figure it out...

If I use a SearchControl with a GeoJSON layer, is it possible to access the found feature and do something with it? I.e what I'd like to do is:

```python
search_control = ipyleaflet.SearchControl(layer=my_layer, property_name='postcode', zoom=12)
m.add_control(search_control)
...
# observe() or something to react to the user finding a single matching entry that would highlight the feature
print(found_feature['properties'])  # or something
```

Of course the crude way to do it would be to iterate through all the features looking for the one with the matching `found_style`, but that really isn't very nice...

Thanks a ton!

sackh:
Thanks, @AnotherKamila for reporting this, as far as I know it is not supported yet for SearchContorl, If I get some time over weekend, I will surely look into it.

PR #707, opened 2020-09-26
Added callback functionality for seached GeoJSON feature in SearchControl

sackh:

@martinRenou - I have added callback for a searched feature in the GeoJSON layer. Please review.
This will address #702

### DIFF
--- a/js/src/controls/SearchControl.js
+++ b/js/src/controls/SearchControl.js
@@ -52,12 +52,11 @@ export class LeafletSearchControlView extends control.LeafletControlView {
   leaflet_events() {
     if (this.model.get('layer') !== null) {
         var found_style = this.model.get('found_style');
-        var self = this;
-        this.obj.on('search:locationfound', function(e) {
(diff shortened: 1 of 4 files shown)


js/src/controls/SearchControl.js
```diff
@@ -52,10 +52,15 @@ export class LeafletSearchControlView extends control.LeafletControlView {
   leaflet_events() {
     if (this.model.get('layer') !== null) {
         var found_style = this.model.get('found_style');
+        var self = this;
         this.obj.on('search:locationfound', function(e) {
```
martinRenou, line 56: You can use an arrow function here `(e) => { ... }`, this will allow you removing `var self = this;` and you can use `this` inside the function.

martinRenou (changes requested):
Thanks!!

js/src/controls/SearchControl.js
sackh: Thanks, I did the change.

Issue #246, opened 2018-12-19
DivIcon layer


martinRenou:
It would be nice to have a `DivIcon` layer. This is a bit related to #207.
https://leafletjs.com/reference-1.3.4.html#divicon

omanges:
@martinRenou  I would like to work on this issue for the JupyterCon sprint, please assign this to me 😊

martinRenou:
Thks!

martinRenou:
Fixed! Thanks @omanges

PR #727, opened 2020-10-17
Added support for DivIcon

omanges:
Added support for DivIcon, related issue https://github.com/jupyter-widgets/ipyleaflet/issues/246


martinRenou:
Neat!!

docs/source/api_reference/divicon.rst
```diff
@@ -0,0 +1,33 @@
...
+----------
+
+=====================    =====================   ===
+Attribute                Default Value           Doc
+=====================    =====================   ===
+html                     ''                      Custom HTML code to put inside the div element,
+bg_pos                   (0, 0)                  Optional relative position of the background, in pixels.
+icon_size                (10, 10)                size icon will be rendered
+icon_anchor              (0, 0)                  anchor point of icon
+class_name               leaflet-div-icon        A custom class name to assign to icon.
```
martinRenou, line 31: I know this is here in some classes already, but I actually think we should not expose any `class_name` property to Python.

The reason is that users don't have a non-hacky way to inject CSS into the page, so I think we should not rely on this property.

ipyleaflet/leaflet.py
```diff
@@ -283,6 +283,56 @@ def _validate_attr(self, proposal):
 
         return value
 
+class DivIcon(UILayer):
+    """DivIcon class.
+
+    Custom lightweight icon for markers that uses a simple <div> element 
+    instead of an image used for markers.
```
martinRenou, line 290: Thanks a ton for this!

martinRenou (commented):
I just have a single comment, apart from that thanks very much for this PR!

docs/source/api_reference/divicon.rst
omanges: Yes right makes sense.

martinRenou:
Thanks!

omanges:
It seems I have now a very good understanding on ipyleaflet 😬

PR #728, opened 2020-10-17
Add WidgetMarker

martinRenou:
This requires updating LeafletJS, we need to be extra careful that everything still works as expected after this.

widgetmarker


martinRenou:
It still needs some CSS work as well.

ipyleaflet/leaflet.py
```diff
@@ -389,6 +389,17 @@ def on_move(self, callback, remove=False):
         self._move_callbacks.register_callback(callback, remove=remove)
 
 
+class WidgetMarker(Marker):
+    _view_name = Unicode('LeafletWidgetMarkerView').tag(sync=True)
```
martinRenou, line 393: Needs docstring

deeplook:
Cool! Would like to see some tiny explanation about which point of the rendered widget the specified location will refer to, like lower-left (or rather South-West?), I'd assume.

martinRenou:
@deeplook I am actually not sure. If you look at the screenshot closely you'll see a small white square at the top-left of the button, I suspect this is the center of the marker but I am really not sure.

martinRenou:
I now position the widget at the center of the marker using CSS

martinRenou:
I might add a `transparent_bg` option like you did @omanges

Issue #716, opened 2020-10-12
WidgetControl transparent background


davidbrochart:
It would be nice to be able to insert a `WidgetControl` on the map without it having a frame and a solid background, so that we can see the map through.

omanges:
@davidbrochart  are you looking for something like this?
![image](https://user-images.githubusercontent.com/30625612/96334192-7a677b00-108c-11eb-8fff-4c285d369e57.png)

davidbrochart:
Yes, that would be awesome!

omanges:
Cool, Can you please assign this issue to me?

omanges:
@davidbrochart can we close this?

martinRenou:
Yes thanks!

PR #732, opened 2020-10-17
To make WidgetControl background transparent

omanges:
To make WidgetControl background transparent as suggested in https://github.com/jupyter-widgets/ipyleaflet/issues/716

### DIFF
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -1395,6 +1395,8 @@ class WidgetControl(Control):
     max_height = Int(default_value=None, allow_none=True).tag(sync=True)
     min_height = Int(default_value=None, allow_none=True).tag(sync=True)
 
+    transparent_bg = Bool(False).tag(sync=True, o=True)
+
 
 class FullScreenControl(Control):
     """FullScreenControl class.
--- a/docs/source/api_reference/widget_control.rst
+++ b/docs/source/api_reference/widget_control.rst
@@ -33,4 +33,5 @@ min_width           None               Min width of the widget (in pixels), if N
 max_width           None               Max width of the widget (in pixels), if None it will respect the content size
 min_height          None               Min height of the widget (in pixels), if None it will respect the content size
 max_height          None               Max height of the widget (in pixels), if None it will respect the content size
+transparent_bg      False              If set to True will make the widget background transparent
 ================    ================   ===
--- a/js/src/controls/WidgetControl.js
+++ b/js/src/controls/WidgetControl.js
@@ -33,7 +33,10 @@ class WidgetControl extends L.Control {
   }
 
   onAdd(map) {
-    this._container = L.DomUtil.create('div', 'leaflet-widgetcontrol');
+    if(this.options.transparentBg)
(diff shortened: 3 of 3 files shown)


ipyleaflet/leaflet.py
```diff
@@ -1395,6 +1395,8 @@ class WidgetControl(Control):
     max_height = Int(default_value=None, allow_none=True).tag(sync=True)
     min_height = Int(default_value=None, allow_none=True).tag(sync=True)
 
+    bg_transparent = Bool(False).tag(sync=True, o=True)
```
martinRenou, line 1398: You can remove `o=True`, this is not an option

davidbrochart, line 1398: Maybe `transparent_bg`?
Why is it an option, unlike the other attributes?

omanges: But will the `bg_transparent` get pulled when`get_options` is called for this 
`this.obj = L.control.widgetcontrol(this.get_options());` ?

davidbrochart:
Thanks a lot @omanges !
Could you rebase on master, so that we don't have all the previous unrelated commits? That would be great for future PRs too.

ipyleaflet/leaflet.py
omanges: Because it needs to get pulled when `get_options` is called over here
`this.obj = L.control.widgetcontrol(this.get_options());`

omanges: So that I can later access it as an option ove here
```
onAdd(map) {
    if(this.options.bgTransparent)
      this._container = L.DomUtil.create('div');
    else
      this._container = L.DomUtil.create('div', 'leaflet-widgetcontrol');
```

martinRenou: This is not an option, you should not use it as an option. But you can access it directly from the model doing:
```javascript
onAdd(map) {
    if(this.model.get('bg_transparent'))
      this._container = L.DomUtil.create('div');
    else
      this._container = L.DomUtil.create('div', 'leaflet-widgetcontrol');
```

omanges: Ok Cool

omanges: But I don’t think the model is accessible in onAdd see below
image

martinRenou: Oh yes, I see now.

I forgot we had a custom Control class for this, then I guess we can make it an official option indeed. But then we might want to do like in Leaflet Controls and officially define it as an option: https://github.com/Leaflet/Leaflet/blob/master/src/control/Control.Zoom.js#L17-L36
This might not be required for it to work but I think it's cleaner.

omanges: How to define it as an option like, because this https://github.com/Leaflet/Leaflet/blob/master/src/control/Control.Zoom.js#L17-L36 syntax is not working with the way currently the class is defined and also  I am not a JS expert 😬

martinRenou: Actually, this seems to only be the default value of the options attribute. Ok let's not bother about this.

PR #733, opened 2020-10-17
[WIP] Add MagnifyingGlass layer

### LINKED ISSUE #720: Magnifying glass
Although it doesn't seem to be maintained, it would be nice to have something like https://github.com/bbecquet/Leaflet.MagnifyingGlass

davidbrochart:
Fixes #720

### DIFF
--- a/js/src/layers/MagnifyingGlass.js
+++ b/js/src/layers/MagnifyingGlass.js
@@ -1,0 +1,47 @@
+// Copyright (c) Jupyter Development Team.
+// Distributed under the terms of the Modified BSD License.
+
+const widgets = require('@jupyter-widgets/base');
+const L = require('../leaflet.js');
+const rasterlayer = require('./RasterLayer.js');
+const layer = require('./Layer.js');
+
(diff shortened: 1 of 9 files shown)


js/src/layers/MagnifyingGlass.js
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  create_obj() {
+    this.layers = [];
+    this.layer_views = new widgets.ViewList(
+      this.add_layer_model,
+      this.remove_layer_view,
+      this
+    );
+    this.layer_views.update(this.model.get('layers'));
```
martinRenou, line 40: This call is async, so the line after your list might be empty.

davidbrochart: Good point, that was it, thanks!

davidbrochart:
Still some work to do with attribute update.

davidbrochart:
An improvement could be to listen to the mouse position on the front-end and to update the `lat_lng` trait accordingly.
Otherwise I think it is ready to be reviewed/merged.

davidbrochart:
image

PR #737, opened 2020-10-18
Add Map.fit_bounds(bounds)

### LINKED ISSUE #96: Add fitBounds method for map
Hi, thanks for the lib. It's awesome!

I havn't found a way to init a map with certain bbox coords. While there is fitBounds() method in Leaflet documentation http://leafletjs.com/reference-1.3.0.html#map-fitbounds

### LINKED ISSUE #529: Feature request: provide access to Leaflet's `fitBounds`
Computing the bounds and maximum zoom level to render all of a layer is a surprisingly complex task, and one that's already implemented in Leaflet:

https://leafletjs.com/reference-1.6.0.html#map-fitbounds

Please consider providing access to this method in the Python API.

davidbrochart:
It is not implemented using Leaflet's [fitBounds method](https://leafletjs.com/reference-1.7.1.html#map-fitbounds) (see issues below), but by iteratively changing the zoom level and checking the bounds (which is the maximum bounds of all map views).

Fixes #96 
Fixes #529

### DIFF
--- a/docs/source/api_reference/map.rst
+++ b/docs/source/api_reference/map.rst
@@ -96,6 +96,7 @@ clear_layers                                                 Remove all layers f
 add_control        Control instance                          Add a new control to the map
 remove_control     Control instance                          Remove a control from the map
 clear_controls                                               Remove all controls from the map
-on_interaction     callable object                           Add a callback on interaction
-save               output file                               Save the map to an HTML file
+on_interaction     Callable object                           Add a callback on interaction
+save               Output file                               Save the map to an HTML file
+fit_bounds         Bounds                                    Set the map so that it contains the given bounds with the maximum zoom level.
 ================   =====================================     ===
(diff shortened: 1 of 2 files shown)


astrojuanlu:
Perhaps something like this could be used to avoid the zoom iteration?

https://github.com/satellogic/telluric/blob/72b0d2dd9e5efe65ba11845fc39c94e83d8394a0/telluric/plotting.py#L56-L79

davidbrochart:
I was thinking about that too, but `mercantile` is spherical mercator only, and ipyleaflet also supports custom projections.

davidbrochart:
Thanks!
Yeah the docs are always lacking. We mention custom projections in the [`crs` attribute of the map](https://ipyleaflet.readthedocs.io/en/latest/api_reference/map.html), but it's really hidden :smile: 
Otherwise there is this notebook: https://github.com/jupyter-widgets/ipyleaflet/blob/master/examples/CustomProjections.ipynb

PR #786, opened 2021-02-04
Fix for Geodata style_callback

### LINKED ISSUE #785: Unable to use GeoData style_callback
GeoJSON layers have a `style_callback` as documented in https://ipyleaflet.readthedocs.io/en/latest/api_reference/geo_json.html, but GeoData subclasses have no such thing documented in https://ipyleaflet.readthedocs.io/en/latest/api_reference/geodata.html. That would be easy to fix, but while using such a GeoData layer I find that the callback seems not to be called or I would know which argument it is called with. A GeoJSON layer callback gets a GeoJSON _feature_, but for GeoData I would rather expect a _row_ of the respective dataframe. But in https://github.com/jupyter-widgets/ipyleaflet/blob/master/ipyleaflet/leaflet.py#L1248 I see no logic to access the dataframe.

sackh:
This will fix #785
@martinRenou  - Please review.


deeplook:
Could there also be a working minimal example like in https://github.com/jupyter-widgets/ipyleaflet/issues/785?

ipyleaflet/leaflet.py
```diff
@@ -1261,10 +1261,16 @@ class GeoData(GeoJSON):
     def __init__(self, **kwargs):
         super(GeoData, self).__init__(**kwargs)
         self.data = self._get_data()
+        self.data = GeoJSON._get_data(self)
```
martinRenou, line 1264: Does it make sense to call `self.data = GeoJSON._get_data(self)` just after `self.data = self._get_data()` ? You overwrites its value making `self._get_data()` useless

martinRenou:
Thanks for the PR!

ipyleaflet/leaflet.py
sackh: `self._get_data()` just converting geodata frame to json and actual logic of `style` and `style_callback` is already implemented in `GeoJSON._get_data` hence reusing that.

deeplook:
What's the status of this PR?

PR #788, opened 2021-02-08
Search control fix

### LINKED ISSUE #787: Search Control doesn't find anything
Hello,
The SearchControl in the maps stopped working since a while, there are no suggestions during the typing and no results after as there used to be.

image

Has there been any other issue on this topic?

Thanks in advance!

sackh:
This will fix #787.
The new version of `leaflet-search` is broken hence pinned the leaflet-search to working version.


martinRenou:
Thks!

AnthonyMacKinnon:
@martinRenou Is there a plan to have a new release that includes this fix? If not, is there a workaround in the meantime?

Issue #823, opened 2021-05-13
Exposing disableClusteringAtZoom option


matyasrada:
Hey,

I had a go with exposing the disableClusteringAtZoom option from leaflet.

I followed the pattern and modified the code on the Python side [ ipyleaflet/leaflet.py ] :

```
class MarkerCluster(Layer):
    """MarkerCluster class.

    A cluster of markers that you can put on the map like other layers.

    Attributes
    ----------
    markers: list, default []
        List of markers to include in the cluster.
    """

    _view_name = Unicode('LeafletMarkerClusterView').tag(sync=True)
    _model_name = Unicode('LeafletMarkerClusterModel').tag(sync=True)

    markers = Tuple().tag(trait=Instance(Marker), sync=True, **widget_serialization)
    
    # Options
    disable_clustering_at_zoom = Int(18).tag(sync=True, o=True)
```

Also modified the code on the js side [ js/src/layers/MarkerCluster.js ]

```
...
const utils = require('../utils');

export class LeafletMarkerClusterModel extends layer.LeafletLayerModel {
  defaults() {
    return {
      ...super.defaults(),
      _view_name: 'LeafletMarkerClusterView',
      _model_name: 'LeafletMarkerClusterModel',
      markers: [],      
      disableClusteringAtZoom: 18,
    };
  }
}
...
  create_obj() {
    var options = this.get_options();
    options.data = this.model.get('data');
    this.obj = L.markerClusterGroup(options);
  }
```

After testing it in a notebook it works perfectly:

```
import geopandas
from ipyleaflet import Map, MarkerCluster, Marker
cities = geopandas.read_file('/path/to/ne_110m_populated_places.shp')
m = Map(center=(42.5, -41.6), zoom=2)
lyr_marker_cluster = MarkerCluster(markers=[Marker(location=geolocation.coords[0][::-1]) for geolocation in cities.geometry], disable_clustering_at_zoom=3)
m.add_layer(lyr_marker_cluster)
m
```

I submitted a PR.

PR #824, opened 2021-05-13
Added disable_marker_clustering_at_zoom option

matyasrada:
Exposing disableClusteringAtZoom option #823.

### DIFF
--- a/js/src/layers/MarkerCluster.js
+++ b/js/src/layers/MarkerCluster.js
@@ -4,6 +4,7 @@
 const widgets = require('@jupyter-widgets/base');
 const L = require('../leaflet.js');
 const layer = require('./Layer.js');
+const utils = require('../utils');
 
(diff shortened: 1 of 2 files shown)


js/src/layers/MarkerCluster.js
```diff
@@ -63,6 +65,10 @@ export class LeafletMarkerClusterView extends layer.LeafletLayerView {
   }
 
   create_obj() {
-    this.obj = L.markerClusterGroup();
+    var options = this.get_options();
+    options.data = this.model.get('data');
```
martinRenou, line 69: What is `this.model.get('data')`?

martinRenou:
Sorry for not looking into this PR before. Thanks!

js/src/layers/MarkerCluster.js
matyasrada: Some remnant code I believe. It can be deleted. my bad

martinRenou:
Thanks!

Issue #508, opened 2020-03-02
Edit external GeoJSON data using DrawControl 


deeplook:
It would be great to be able to use a DrawControl not only for drawing new shapes, but also for loading and editing existing ones in order to save them as well. I've done some experiments along these lines, but it seems like the existing DrawControl is too limited to allow for this. Here some sample snippets (with no effect):

```python
import requests
from ipyleaflet import Map, DrawControl

m = Map(center=[41.0462, 20.1388], zoom=6)
dc = DrawControl()
m += dc
display(m)
url = 'https://github.com/jupyter-widgets/ipyleaflet/raw/master/examples/europe_110.geo.json'
albania_feature = requests.get(url).json()["features"][0]
dc.last_draw = albania_feature  # has no effect
```

```python
import requests
from ipyleaflet import Map, DrawControl, GeoJSON

m = Map(center=[41.0462, 20.1388], zoom=6)
dc = DrawControl()
m += dc
url = 'https://github.com/jupyter-widgets/ipyleaflet/raw/master/examples/europe_110.geo.json'
albania_feature = requests.get(url).json()["features"][0]
m += GeoJSON(data=albania_feature)  # shows Albania as expected
display(m)
```

I could also not find any handle in the DrawControl instance to any “drawn” shape other than the last one. So it could well be that the wrapped Leaflet DrawControl is lacking this feature. This could also be related to #358.

exospatial:
It used to be possible using the layer attribute like this:
```
dc = DrawControl()
dc.layer = GeoJSON(data=albania_feature)
```
The reasons for the layer attribute being removed are listed in this issue #68
I did use the feature and would be happy to help implement a replacement but could do with some pointers from someone familiar with the internals of the ipyleaflet

dbergh:
I added PR #826 addressing this issue.

giswqs:
I have implemented this in [leafmap](https://github.com/giswqs/leafmap/). You can now edit existing vector data interactively. 

https://user-images.githubusercontent.com/5016453/151687188-bdf136d0-d70f-4acb-86b6-cff580971d65.mp4

Issue #817, opened 2021-04-26
Editable vertices on Paths/Polylines/Geojson/Geodata


AndrewAnnex:
I want to be able to load a geopandas dataframe into a ipyleaflet map, and be able to move/edit/drag vertices of the loaded geometries (in my case linestrings) on the ipyleaflet map. I also want to then take actions using the callbacks so I can have python code handle the specific geometry that was edited for external python processes to act on and so on (I edited the vertex of a geometry in the dataframe, I would want to know which row/geometry/ what the new geometry is/ etc).

I know this is possible with individual markers, but for more complicated geometries  it appears it may not be at the moment due to changes noted in #508 #789 #68.

Is there a way to get this functionality with the existing most recent release? If the change is trivial (or somewhat) I am happy to contribute.

dbergh:
I added PR #826 to address #508. It does not give you quite what you want (using this you'd get an update of the whole geometry, but individual features are not tracked). Maybe you can use it for some kind of work around...

AndrewAnnex:
thanks @dbergh, I actually found out the inner portion of my code was fast enough such that updating the whole geometry wouldn't be an issue. I ended up continuing my work in a qgis plugin, but since it's all python it shouldn't be too much work to migrate that effort back to ipyleaflet once it's merged and released. Closing the issue

Issue #789, opened 2021-02-11
DrawControl imported polygon


eironman3:
Hello guys, 

Is there any way to import a polygon shape into the map (from my database) and edit it using DrawControl?
I manage to create new polygons, export them, etc but I couldn't find a way to reshape an imported geometry.

I have been looking around and it looks like there was a discussion here #508. 
Do you guys know if there is a workaround for this?

Thanks a lot.

I really appreciate your help.

martinRenou:
Hey.

There is a `transform` attribute that you can set to True on your `Polygon` instance, making it editable. Does that fit your needs?

eironman3:
hi Martin, 

Thanks a lot for your answer. 
The transform option only allows you to change the size of the polygon + rotate it but I can't add/delete any additional points.

martinRenou:
Indeed. There is currently no possible way to do that I guess. We might need to improve the DrawControl to provide this ability.

dbergh:
I added PR #826 which should address this issue.

giswqs:
The `data` attribute of the DrawControl implemented by @dbergh should resolve this issue. 

Alternatively, you can also use [leafmap](https://github.com/giswqs/leafmap/) to edit existing vector data interactively. 

https://user-images.githubusercontent.com/5016453/151687323-717d25bf-0c83-460c-bfb0-ab95c0e89671.mp4

Issue #68, opened 2017-07-20
How to remove the Layers when i created polygon use DrawControl?


tracegyx:
I know that there is a button on map and use it to delete the layers, but i want delete the layers use code, so what should i do?
![e9b495c8-e6ee-499c-95ff-0aa323e40d12](https://user-images.githubusercontent.com/21274507/28400980-58da5550-6d49-11e7-9a0a-843d0870b39d.png)

GriffinLane:
I too am looking for a feature that will allow me to delete previous rectangles and/or limit their quantity.  Bump +1

martinRenou:
It's actually not that easy to implement. First, there is a bug in ipyleaflet, the `FeatureGroup` class should inherit from `LayerGroup` class, like in `Leaflet`: https://github.com/Leaflet/Leaflet/blob/e5a3d5ecc360af77e291451172c60d4e3807cf72/src/layer/FeatureGroup.js#L26 so that we could have access to `add_layer`, `remove_layer` methods.
I open a PR for this first issue, but even after this fix, it will still not be possible to fix your issue.

The `DrawControl` widget owns a `layer` attribute which is a `FeatureGroup` instance, you can add new layers (rectangle, markers...) to this `layer` attribute from the python side and it works fine. 

BUT if a new layer has been added to this featuregroup from the javascript side (for instance when drawing a rectangle with your mouse), the list of layers won't be synchronized anymore between the python and javascript. So you won't have access to new added layers from the python side, which is why it's not possible with the current implementation of `FeatureGroup` to implement your request.

However, I found a workaround so that it's possible to remove drawings, hiding the complexity of the `FeatureGroup` from the python side, I'll open a PR soon.

martinRenou:
My PR removes the `layer` attribute of `DrawControl`, but as `FeatureGroup` was not working, I think it's safe to say that nobody was using it

PR #826, opened 2021-05-24
Added data attribute to DrawControl

dbergh:
## Description

A data attribute is added to the model of the DrawControl widget. This allows the editable data layer to be changed programmatically from python. In addition, the attribute is automatically synced on each edit. The data attribute contains a
list dicts representing GeoJson objects. These are automatically converted into layers on the javascript side. It is possible to provide styling information by setting the property.style attribute.

The solution is not perfect as it does not give full control of the layers in the editable feature collection, but it does the job supporting many important use cases. In particular, it should solve issue #508.

This solution renders the DrawControl methods `clear`, `clear_polygons` obsolete.

## Demonstration code
```python

m = Map(center=[55.71937583528782, 13.13922374405463], zoom=10)
dc = DrawControl()
m += dc
dc.data = [{'type': 'Feature',
  'properties': {'style': {'stroke': True,
    'color': '#3388ff',
    'weight': 4,
    'opacity': 0.5,
    'fill': True,
    'fillColor': None,
    'fillOpacity': 0.2,
    'clickable': True}},
  'geometry': {'type': 'Polygon',
   'coordinates': [[[13.151797, 55.730145],
     [13.247946, 55.727439],
     [13.244512, 55.681398],
     [13.155918, 55.68372],
     [13.151797, 55.730145]]]}}]
display(m)
```

## References
This should solve issue #508

## Know issues
The way the style attribute is treated is a little bit sketchy. If you edit a provided shape, and then read the data again, typically extra style attributes are added. I am not sure how to fix this in the best way.

### DIFF
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -1661,6 +1661,9 @@ class DrawControl(Control):
     edit = Bool(True).tag(sync=True)
     remove = Bool(True).tag(sync=True)
 
+    # Layer data
+    data = List().tag(sync=True)
+
     last_draw = Dict({
         'type': 'Feature',
         'geometry': None
--- a/js/src/controls/DrawControl.js
+++ b/js/src/controls/DrawControl.js
@@ -17,6 +17,7 @@ export class LeafletDrawControlModel extends control.LeafletControlModel {
       circlemarker: {},
       rectangle: {},
       marker: {},
+      data: [],
       edit: true,
       remove: true
     };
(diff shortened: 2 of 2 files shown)


dbergh:
This is also related to issues #789 #68 #817

martinRenou:
Thanks!

martinRenou:
> This solution renders the DrawControl methods clear, clear_polygons obsolete

Indeed. We should get rid of those at some point then :) Maybe in another PR

dbergh:
I just realized that although my solution works ok for polygons, it does not work well for other shapes. Since I just used a GeoJSON-like object in the data-attribute, there is no way to distinguish between for instance rectangles and general polygons. Perhaps we should also add properties for shape type and other parameters (as for instance radius for circles). Any suggestions on how to do this?

PR #835, opened 2021-06-05
Corrected fit_bounds docs

giswqs:
The `fit_bounds` method documentation `[[south, east], [north, west]]` is incorrect. It should be `[[south, west], [north, east]]`


davidbrochart:
Thanks!

PR #836, opened 2021-06-06
Update related projects

davidbrochart:
@giswqs I added leafmap and updated the URL of geemap.
Let me know if that's fine!


giswqs:
That's great. Thank you!

PR #842, opened 2021-06-10
Add --overwrite option for a dev install with JLab

maximlt:
Since without it a ValueError is raised (ref https://github.com/jupyter-widgets/widget-cookiecutter/issues/88)


davidbrochart:
Thanks!

PR #846, opened 2021-06-17
Remove shapely and branca dependencies

### LINKED ISSUE #769: ipyleaflet v0.13.4 import error on Windows 10
The pip installation works fine on Windows 10. However,  `import ipyleaflet` is throwing an error due to the missing `geos_c.dll`. This is likely caused by the recently added dependency of shapely.

martinRenou:
This will allow using ipyleaflet in jupyterlite (as shapely is not installable with micropip).

Fix #769

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -4,3 +4,5 @@ scipy
 netcdf4
 geopandas
 xarray
+branca
+shapely
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -19,9 +19,6 @@ from traitlets import (
     link, observe, default, validate, TraitError, Union, Any
 )
 
-from branca.colormap import linear, ColorMap
-from shapely import geometry, wkt
-
 from ._version import EXTENSION_VERSION
 
 from .projections import projections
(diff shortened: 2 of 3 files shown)


davidbrochart:
Do you know if it's the last one? If yes, then maybe we can think of an alternative to branca.

martinRenou:
I honestly thought of removing branca the day I added it ahah. I think we should use bqscales so that we can share a colormap between e.g. ipyleaflet and bqplot.

But bqscales needs more features to catch up on branca.


Actually folium could be installed on jupyterlite and it uses branca, they workaround it by installing jinja2 and branca this way:
```python
await micropip.install("https://files.pythonhosted.org/packages/61/1f/570b0615c452265d57e4114e633231d6cd9b9d275256778a675681e4f711/branca-0.4.2-py3-none-any.whl")
await micropip.install('https://files.pythonhosted.org/packages/80/21/ae597efc7ed8caaa43fb35062288baaf99a7d43ff0cf66452ddf47604ee6/Jinja2-3.0.1-py3-none-any.whl')

```

giswqs:
This is nice! Removing shapely dependency will also fix the Windows pip installation issue. Can't wait to try out ipyleaflet in JupyterLite. I have been waiting forever for Colab to support ipyleaflet. Now JupyterLite comes to rescue. Thank you!

PR #853, opened 2021-07-12
Added a new Map option "prefer_canvas".

### LINKED ISSUE #852: How to use the canvas renderer for plotting larger GeoJSON/GeoData dataset
I have a modest GeoJSON/GeoData dataset of point features. Currently I use the `point_style` attribute of the `GeoData` constructor to plot the points as circles. This renders the circles as SVG features which has performance issues with larger feature counts.
The Leaflet allows us to render using the canvas renderer in such cases, see https://leafletjs.com/reference-1.7.1.html#map-prefercanvas or https://stackoverflow.com/questions/43015854/large-dataset-of-markers-or-dots-in-leaflet.

**How can we force the Leaflet map to use the canvas renderer in the Jupyter notebook?**

My current code:
```
   m = ipyl.Map()
    layer = ipyl.GeoData(geo_dataframe = data_to_show,
                        point_style={'radius': 1, 'color': 'amber'})
    m.add_layer(layer)
    
```

ondrejsv:
Resolves #852 by introducing a new `Map` option `prefer_canvas` which passes down to the Leaflet map instance.

### DIFF
--- a/examples/CanvasRenderer.ipynb
+++ b/examples/CanvasRenderer.ipynb
@@ -1,0 +1,90 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "24ede0d7-88e1-4955-9094-e10a9aebe5e7",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from ipyleaflet import Map, GeoData\n",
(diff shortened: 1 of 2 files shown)


martinRenou:
Great thanks!

PR #856, opened 2021-08-06
Fixed typo

giswqs:
Fixed several topos using [codespell](https://github.com/codespell-project/codespell).

### DIFF
--- a/examples/LegendControl.ipynb
+++ b/examples/LegendControl.ipynb
@@ -155,7 +155,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "### positioning"
+    "### Positioning"
    ]
   },
   {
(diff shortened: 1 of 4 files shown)


examples/LegendControl.ipynb
```diff
@@ -155,7 +155,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "### Positionning"
+    "### positioning"
```
davidbrochart, line 158:
```suggestion
    "### Positioning"
```

ipyleaflet/leaflet.py
```diff
@@ -1841,11 +1841,11 @@ def legends(self, legends):
         self.legend = legends
 
     @property
-    def positionning(self):
+    def positioning(self):
```
davidbrochart, line 1844: Maybe we could keep the old `positionning` and make it call the new `positioning`, for compatibility reasons?

giswqs: Just added `positionning`.

davidbrochart:
Thanks!

PR #857, opened 2021-08-09
Added support for xyzservices tile providers as basemaps

### LINKED ISSUE #854: Refactor basemap module on top of xyzservices
Hi all,

I'd like to start a focused discussion as a follow-up of https://github.com/geopandas/contextily/issues/153 and https://github.com/jupyter-widgets/ipyleaflet/issues/741. 

We have built a new package called [`xyzservices`](https://github.com/geopandas/xyzservices) to serve as a unified place for XYZ tile management. The idea behind this is that other packages will depend on that instead of having their own system.

In the case of `ipyleaflet`, that means refactoring `basemaps` module in a way that does not contain any hard-coded data and only wraps what ``xyzservices` provides.

From a quick look, it could take the form of replacing your `Bunch` with `xyzservices.providers` Bunch and of small changes in https://github.com/jupyter-widgets/ipyleaflet/blob/c969f79da419aef39f63e6745d0f234493280f67/ipyleaflet/leaflet.py#L36 to match the `xyzservices` API (to get the proper form of URL (xref geopandas/xyzservices/pull/60) and attribution).

`xyzservices` itself does not have any dependencies which makes it a non-problematic dependency.

Some of you also mentioned you wanted the JSON stored in `share`. That also comes with `xyzservices` if you have a use case for that.

I am happy to do any changes in `xyzservices` needed to make it as useful for you as possible if that allows centralising maintenance of providers in one place.

What are your thoughts on this?

sackh:

This PR will resolve #854 
I have tested all basemaps.
The below basemaps will not work at the moment as they are not defined in [xyzservices](https://github.com/geopandas/xyzservices) providers:

Map(basemap=basemaps.OpenStreetMap.BlackAndWhite, center=center, zoom=zoom) 
Map(basemap=basemaps.Gaode.Normal, center=center, zoom=zoom)  
Map(basemap=basemaps.Gaode.Satellite, center=center, zoom=zoom) 
Map(basemap=basemaps.NASAGIBS.ModisTerraBands721CR, center=center, zoom=zoom) 
Map(basemap=basemaps.NASAGIBS.ModisAquaTrueColorCR, center=center, zoom=zoom)
Map(basemap=basemaps.NASAGIBS.ModisAquaBands721CR, center=center, zoom=zoom) 
Map(basemap=basemaps.NASAGIBS.ViirsTrueColorCR, center=center, zoom=zoom) 
Map(basemap=basemaps.Strava.Ride, center=center, zoom=zoom) 
Map(basemap=basemaps.Strava.Run, center=center, zoom=zoom)
Map(basemap=basemaps.Strava.Water, center=center, zoom=zoom)
Map(basemap=basemaps.Strava.Winter, center=center, zoom=zoom)


ipyleaflet/leaflet.py
```diff
@@ -2135,7 +2128,7 @@ def add_layer(self, layer):
         layer: Layer instance
             The new layer to add.
         """
-        if isinstance(layer, dict):
+        if isinstance(layer, (dict, xyzservices.lib.TileProvider)):
```
martinfleis, line 2131: Just FYI, `xyzservices.lib.TileProvider` is a subclass of `dict` so the original condition would work as well.

martinfleis (commented):
This is exactly the kind of PR we were hoping for when we started building `xyzservices` :):
Screenshot 2021-08-09 at 19 48 33 

The missing providers will be added in geopandas/xyzservices#83 and we can cut a release shortly after that if that helps. I am just not sure if Strava works at all.

Thanks a lot for playing with us :).

ipyleaflet/leaflet.py
sackh: done

martinRenou:
Thanks for doing this!

martinfleis:
I have released `xyzservices 2021.08.1` which includes those tiles listed above as missing. That should fix your CI as well. It should show up on conda-forge in a few hours.

davidbrochart (approved):

ipyleaflet/leaflet.py
```diff
@@ -32,35 +34,28 @@
...
-    from datetime import date, timedelta
-
-    if day == 'yesterday':
-        yesterday = date.today() - timedelta(1)
-        day = yesterday.strftime('%Y-%m-%d')
-
-    url = basemap.get('url', '')
-    if url.count('%'):
-        url = url % day
-
```
martinRenou, line 58: This will be a breaking change for people who were manually doing the string formatting, for example this Notebook does not work anymore: https://github.com/jupyter-widgets/tutorial/blob/master/notebooks/extras/ipyleaflet.ipynb.

(just a note that we shouldn't make a patch release with this)

sackh: I think string formatting with different placeholders, can not be made backward compatible.

martinRenou:
Rebased

martinRenou:
Thanks a lot for this PR!

The docs don't build, but it's not due to this PR so let's ignore it.

Issue #750, opened 2020-11-24
add info about resizing to docs?


raybellwaves:
I was looking how to this and I came across

https://github.com/jupyter-widgets/ipyleaflet/issues/345

Could add a little section in the docs on this e.g.

```
from ipyleaflet import Map, basemaps

center = [38.128, 2.588]
zoom = 5

m = Map(basemap=basemaps.OpenStreetMap.Mapnik, center=center, zoom=zoom)
m.layout.height = '600px'
m.layout.width = '600px'
```

martinRenou:
Indeed that would be helpful! Would you like to open a PR and get credits for this?

raybellwaves:
Any thoughts on where this could go?

The docs are nicely structured.

martinRenou:
I am tempted to suggest a new section in the map documentation: https://github.com/jupyter-widgets/ipyleaflet/blob/master/docs/source/api_reference/map.rst

PR #868, opened 2021-09-11
Add doc section for Map on basic layout options. Issue #750

jmp75:
How to customise the layout is not obvious to new users not already familiar with ipywidgets. Screen sizes these days are such the default layout appears very horizontal and users appear often inclined to change the default height. The first experience to find out how can be smoothed a bit via documentation.


martinRenou:
Thanks! The docs appear to fail building but it's not due to this PR

Issue #865, opened 2021-09-02
map not visible


maessery:
hi, i was looking at your work on https://ipyleaflet.readthedocs.io/en/latest/api_reference/basemaps.html#
 with basemaps, specifically, and noticed all your Strava maps were greyed out. There was a note box that suggested reaching out on  here if that situation occured on the site.

As a side note, im interested to see what it produces! Is it a way of plotting streets and drivable routes?

davidbrochart:
Hi, we depend on `xyzservices`, and it seems [the Strava server is currently down](https://github.com/geopandas/xyzservices/blob/f6f5e8640596721400af5c2bdbefe2fc150dae72/CHANGELOG.md#xyzservices-2021081-august-12-2021).

justb4:
From what I see the problem is just a typo double slash, '//tiles/' in the Strava URLs on [basemaps.py#199 and further](https://github.com/jupyter-widgets/ipyleaflet/blob/master/ipyleaflet/basemaps.py#L199):

```
    Strava=Bunch(
        All=dict(
            url='https://heatmap-external-a.strava.com//tiles/all/hot/{z}/{x}/{y}.png?v=19',
            max_zoom=15,
            attribution='Map tiles by <a href="https://labs.strava.com/heatmap">Strava 2017</a>',
            name='Strava.All'
        ),
 .
.

```
So https://heatmap-external-a.strava.com/tiles/all/hot/10/163/393.png will render but 
https://heatmap-external-a.strava.com//tiles/all/hot/10/163/393.png not.

justb4:
And https://github.com/geopandas/xyzservices/pull/85 (the more `xyzservices` becomes relevant!).

martinRenou:
Thanks for reporting. So it might only be a matter of updating the docs. We maybe need to update the stable branch.

martinRenou:
Thanks for reporting. So it might only be a matter of updating the docs. We maybe need to update the stable branch.

martinRenou:
Rebuilding now

justb4:
Not too familiar how xyzservices is used, thought it was WIP via PR #857.
To me it seems that only
https://github.com/jupyter-widgets/ipyleaflet/blob/master/ipyleaflet/basemaps.py#L199 needs updating...

martinRenou:
Indeed. I thought the xyzservices PR was merged and released, but it is still open https://github.com/jupyter-widgets/ipyleaflet/pull/857.

[...]

PR #870, opened 2021-09-12
Fix for #865 : remove double slashes from Strava tile-URLs

justb4:
Simply replace `'...com//tiles/...'` by `'...com//tiles/...'` in Strava URLs


martinRenou:
Thanks!
\

Issue #811, opened 2021-03-30
LayerGroup not exported when saving to html


Kuifje02:
Hi ! 

I have noticed that if I use [Layer Groups](https://ipyleaflet.readthedocs.io/en/latest/api_reference/layer_group.html?highlight=layer%20group) and export as html with `m.save('my_map.html', title='My Map')`, the export fails (no layers appear in the html file).

thanks !

obdura:
Same problem here. Layer Control and layers are not displayed on exported html page.

PR #873, opened 2021-10-01
Declare LayerGroup view_module and model_module

philippjfr:
The `LayerGroup` was not inheriting the `_view_module` and `_model_module` from the super, which meant that the view and model could not be resolved in exported contexts. I'm guessing it was intentional not to call `...super.defaults()` to avoid inheriting a bunch of other attributes from the `LeafletLayerModel` but I'd love a confirmation of that from you @martinRenou.

Fixes https://github.com/jupyter-widgets/ipyleaflet/issues/811
Fixes https://github.com/jupyter-widgets/ipyleaflet/issues/761

### DIFF
--- a/js/src/layers/LayerGroup.js
+++ b/js/src/layers/LayerGroup.js
@@ -10,8 +10,10 @@ export class LeafletLayerGroupModel extends layer.LeafletLayerModel {
     return {
       _view_name: 'LeafletLayerGroupView',
       _model_name: 'LeafletLayerGroupModel',
-      _view_module: 'jupyter-leaflet',
       _model_module: 'jupyter-leaflet',
+      _model_module_version: null,
+      _view_module: 'jupyter-leaflet',
+      _view_module_version: null,
       layers: []
     };
   }
--- a/js/src/layers/LayerGroup.js
+++ b/js/src/layers/LayerGroup.js
@@ -10,6 +10,8 @@ export class LeafletLayerGroupModel extends layer.LeafletLayerModel {
     return {
       _view_name: 'LeafletLayerGroupView',
       _model_name: 'LeafletLayerGroupModel',
+      _view_module: 'jupyter-leaflet',
+      _model_module: 'jupyter-leaflet',
       layers: []
     };
   }


js/src/layers/LayerGroup.js
```diff
@@ -10,6 +10,8 @@ export class LeafletLayerGroupModel extends layer.LeafletLayerModel {
     return {
       _view_name: 'LeafletLayerGroupView',
       _model_name: 'LeafletLayerGroupModel',
+      _view_module: 'jupyter-leaflet',
+      _model_module: 'jupyter-leaflet',
```
philippjfr, line 14: I guess this may not actually be enough, do we also need the `_view_module_version` and the `_model_module_version`?

martinRenou: Yes we might need them :)

martinRenou:
Thanks! I wonder why/how it was working without this Oo

js/src/layers/LayerGroup.js
philippjfr: Do I just initialize them to null?

philippjfr:
>Thanks! I wonder why/how it was working without this Oo

I think this is only really critical in the static html renderer where these are used to fetch the library from NPM.

Anything else I can do here? Not sure what to initialize the version values to.

js/src/layers/LayerGroup.js
```diff
@@ -10,6 +10,10 @@ export class LeafletLayerGroupModel extends layer.LeafletLayerModel {
     return {
       _view_name: 'LeafletLayerGroupView',
```
martinRenou, line 11: Sorry for not looking at this more deeply the other time. This is actually missing a call to the `super`'s `defaults` method:

```javascript
...super.defaults(),
```

Similar to here: https://github.com/jupyter-widgets/ipyleaflet/blob/master/js/src/layers/AntPath.js#L11

This will merge the superclass defaults attribute which contains the proper module names and versions. 

`LayerGroup` is actually not the only one missing this, `ImageOverlay` also misses it (and maybe others).

martinRenou (changes requested):
Sorry for the delay in reviewing this

martinRenou:
I just opened #894 that replaces this PR. Thank you @philippjfr for catching this issue!

philippjfr:
Thanks @martinRenou, sorry I didn't get a chance to follow up.

martinRenou:
No worries. I should have properly reviewed your PR the first time to not waste everyone's time, totally my fault.

Issue #881, opened 2021-11-04
Very slow CircleMarker creation (?)


mangecoeur:
I'm creating a map with a "medium" number of CircleMarkers in JupyterLab and having some performance issues, it seems that just creating a CircleMarker is very slow, e.g:
```python
%%time
from ipyleaflet import CircleMarker
for i in range(1000):
    c = CircleMarker()
```
Takes 1.4seconds, I would not expect creating 1000 objects to be that slow (would expect timings in the milli or micro second range). Is there an explanation for this or a workaround?

I'm using latest JupyterLab via JupyterHub running on an Ubuntu server, the client is Firefox 94 on Windows 10.

mangecoeur:
Deeper investigation suggests that this happens because every CircleMarker call instantiates a Widget with bidirectional comms between the kernel and the client. In my case I don't need this. It seems in Leaflet you can instantiate a MarkerCluster from geojson point data directly by adding it as a layer in MarkerCluster, which would be ideal since the geojson sync is very fast. However MarkerCluster widget side doesn't seem to allow this (?)

martinRenou:
> call instantiates a Widget with bidirectional comms between the kernel and the client. In my case I don't need this

Would using a `GeoJSON` layer instead be an option?

mangecoeur:
> Would using a GeoJSON layer instead be an option?

The problem is I want keep the cluster behavior of the MarkerCluster. In Leaflet it's possible to add a GeoJSON layer of points to a MarkerCluster, this would be exactly what I need. However, it is not possible in ipyleaflet to set the markers attribute to be a GeoJSON layer (trait type expects `tuple(Marker)`). I don't know if it would be possible to change this without a deeper re-working of MarkerCluster - I would love any hints on this point.

martinRenou:
I see. Then I guess this would need a rework on the `MarkerCluster` logic in ipyleaflet as you suggested.

You're welcome to open a PR for this if you want to contribute :) I'd be happy to guide you a bit.

mangecoeur:
Ok, I thought that might be the case. I think I could attempt this since it's really bugging me (I optimised the hell out of my calculation only to have it hang for 3 seconds while loading little circles ;)  I've had a glance at the widget doc but any pointer on where to start are very welcome!

martinRenou:
[...]

PR #882, opened 2021-11-08
Support GeoJson layer in MarkerCluster

mangecoeur:
Initial PR (needs review, maybe tests...) addessing #881

Made it possible to use a GeoJson layer as the data source for Marker Cluster, this makes it much faster when creating large numbers of markers (but not having bidirectional sync on individual Marker widgets).

Implemented by copying the behaviour of LayerGroup, it will simply attempt to add the model of whatever is provided in the marker list to the marker group: **this allows to mix geojson and Marker elements**.

### DIFF
--- a/examples/MarkerCluster-GeoJson.ipynb
+++ b/examples/MarkerCluster-GeoJson.ipynb
@@ -30,7 +30,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 4,
+   "execution_count": 30,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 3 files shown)


martinRenou:
That's very neat! Thanks a lot! And thanks for using a view list, when looking at the code yesterday I thought we should do that.

I will review more deeply and test it today 😊

mangecoeur:
Thanks! Would be great to have this enabled.

examples/MarkerCluster-GeoJson.ipynb
```diff
@@ -0,0 +1,104 @@
+{
```
martinRenou, line 1: Could you clean the Notebook outputs?

martinRenou (commented):
Thanks a lot! Works like a charm :)

It's a really nice improvement, I've just got two small comments and it's good to go

mangecoeur:
Great thanks, just pushed a commit to remove console.log and clean the nb outputs.

martinRenou:
Thanks!

PR #905, opened 2021-12-16
Test `ipyleaflet` widgets with `galata`

trungleduc:


davidbrochart:
Do all the example notebooks need to be committed?

trungleduc:
> Do all the example notebooks need to be committed?

No , it's my bad

trungleduc:
Hi all, this PR is ready for review.

davidbrochart (approved):

martinRenou:
We have a slightly different setup in bqplot where we also test that updating widget's traits have the correct impact on the widget view.

I think it makes sense to do something similar for ipyleaflet, this would allow to run a cell that changes a marker position `marker.location = (53., 45.)` and take a screenshot of the map again, checking that the marker was updated as expected.

martinRenou:
This can definitely be done in a separate PR though. That's already a great start, thank you for doing this!

trungleduc:
> We have a slightly different setup in bqplot where we also test that updating widget's traits have the correct impact on the widget view.
> 
> See https://github.com/bqplot/bqplot/blob/master/ui-tests/tests/bqplot.test.ts
> 
> I think it makes sense to do something similar for ipyleaflet, this would allow to run a cell that changes a marker position `marker.location = (53., 45.)` and take a screenshot of the map again, checking that the marker was updated as expected.

indeed, I just add the simplest tests which try to render the available widgets of `ipyleaflet`. We can iterate to add more complicated tests after.

martinRenou:
🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

@deeplook Actual tests in ipyleaflet!!!

PR #907, opened 2021-12-20
Add bounds attribute to TileLayer

### LINKED ISSUE #888: Constrain bounds of TileLayer
Would it be possible to add a property to the `TileLayer` that will constrain its bounds? 

The lack of constraint poses a problem when accessing a tile server I have that serves individual rasters (small coverage, not global). The `TileLayer` continues to make requests from my tile server anytime I zoom out/view outside the data bounds, overloading my tile server which is just returning a bunch of transparent PNGs.

The logic I'm looking for:

```py
t = TileLayer(
    url="https://mytileserver.com/tiles/{z}/{x}/{y}.png?projection=EPSG:3857",
    # Do not request tile outside of these bounds (bottom, top, left, right)
    bounds=[37.40679, 38.080379, -122.759396, -122.065103],
)
```

From what I can tell, the leaflet JS tileLayer does have an option for `bounds` that is inherited from gridLayer but this does not seem to be wrapped in `ipyleaflet`. Ref https://leafletjs.com/reference.html#tilelayer

Example data:

Screen Shot 2021-11-20 at 8 33 19 PM

-------

Additional context:

- https://github.com/banesullivan/flask-tileserver / https://github.com/jupyter-widgets/ipyleaflet/issues/234#issuecomment-974716145
- https://github.com/ResonantGeoData/ResonantGeoData/pull/603

davidbrochart:
Closes #888


davidbrochart:
Not sure about the UI test failure, any idea @trungleduc ?

trungleduc:
it looks like  some texts on the map have changed

- Expected map
SplitMapControl-expected

- Rendered map
SplitMapControl-actual

- Diff

SplitMapControl-diff

trungleduc:
You can update `SplitMapControl-linux.png` file with the rendered map to make the tests pass, we can remove this test if the screenshot keeps changing.

davidbrochart:
Thanks Trung, I uploaded the rendered snapshot. Let's hope the base map doesn't change too often.

martinRenou:
Thanks! You also need to change the Javascript model

davidbrochart:
Thanks! Interestingly, it was already working without it.

martinRenou:
Yes, it's only used in some cases when the widget is embedded.

davidbrochart:
@trungleduc where are the UI test assets uploaded? I cannot find them.

trungleduc:
https://github.com/jupyter-widgets/ipyleaflet/actions/runs/1605786157, Artifacts section

trungleduc:
I think we should update the `SplitMapControl.ipynb` notebook to move the map somewhere else

davidbrochart:
I removed the update of the rendered split map, tests are passing now.
I'm wondering if the tiles are not still loading when the snapshot is made, e.g. maybe we change the zoom and we don't wait enough?

martinRenou:
Did you see there was a very nice splitmap widget on Github for checking the splitmap screenshot?


https://user-images.githubusercontent.com/21197331/146900522-6819aa81-4123-4da6-8f2d-27fd67fbeb28.mp4

Issue #912, opened 2022-01-13
Opening a popup on clicking a GeoJSON object


PROgram52bc:
From the documentation on [popup](https://ipyleaflet.readthedocs.io/en/latest/api_reference/popup.html), it seems like popups can only be opened interactively with the marker element. 

I wonder if it can also be opened via a custom callback function attached to a `GeoJSON.on_click` method, adding something like `popup.open()` or `popup.close()` to the `Popup` class.

Seems to be related to #239. Any suggestions on how I should proceed from here?

martinRenou:
> adding something like popup.open() or popup.close() to the Popup class

That sounds like a good idea :)

If you would like to implement this feature, I guess it will be similar to the `redraw` function the `TileLayer` class provides: https://github.com/jupyter-widgets/ipyleaflet/blob/master/ipyleaflet/leaflet.py#L589-L595
It sends a comm message asking for a `redraw`, and the front-end handles it here: https://github.com/jupyter-widgets/ipyleaflet/blob/3273696bfe73d051fc7864f6731c8ed2198e6f68/js/src/layers/TileLayer.js#L76-L80
Don't hesitate to ask for more guidance :) 
Opening the popup in JS will look like the following I guess:
```javascript
this.map_view.el.openPopup(this.el);  // "this.el" is the Popup object at that point
```

PROgram52bc:
Thanks! I will look into this sometime next week.

PROgram52bc:
I installed the development environment following [this document](https://github.com/jupyter-widgets/ipyleaflet/blob/master/docs/source/installation.rst), but now the issue is that I cannot see changes that I made on the javascript side (e.g. adding a `console.log` statement to `js/src/leaflet.js`). I have a separate console running `npm run watch`, but webpack seems not to have seen the change I have made. Any suggestions? Is that the right way to set up the dev environment?

martinRenou:
Are you developing on Jupyter Notebook or JupyterLab?

PROgram52bc:
> Are you developing on Jupyter Notebook or JupyterLab?

I'm using jupyter notebook

martinRenou:
Did you run the command:
```
jupyter nbextension install --py --symlink --sys-prefix --overwrite ipyleaflet
jupyter nbextension enable --py --sys-prefix ipyleaflet
```
?

The dev docs is missing the `--overwrite` option that you might need

martinRenou:
[...]

Issue #239, opened 2018-12-10
Open/closed status of popups


davidbrochart:
Is there a way to know in Python if a Popup has been closed in the front-end?

martinRenou:
There's a [method](https://leafletjs.com/reference-1.3.4.html#popup-isopen) on the front-end to retrieve the state of the popup. So I guess we can try to add a new trait to the `Popup` Python class.

davidbrochart:
Yes it looks good.

davidbrochart:
`isOpen()` has to be called explicitly in JS, and its result returned back to Python. I would like to have a Python Popup trait `is_open` that would do that (synchronously), but I can't figure out a way to do it. Am I missing something or is it not that trivial?

martinRenou:
Ok, so `isOpen` wasn't a good choice actually. You can listen to the [popupopen](https://leafletjs.com/reference-1.3.4.html#popup-popupopen) and popupclose events for the Javascript->Python update.
For the Python->Javascript update, you can maybe use `this.map_view..obj.openPopup(this.obj)` in the `PopupView` class on the Javascript side. And I guess there is also a `closePopup`

martinRenou:
There is also a [openPopup](https://github.com/Leaflet/Leaflet/blob/55755f41bedc47a43ee451a3b3f454ddb5f05f2c/src/layer/Popup.js#L420) and `closePopup` for each layer instance. Which could be usefull for the [popup](https://github.com/jupyter-widgets/ipyleaflet/blob/master/ipyleaflet/leaflet.py#L72) attributre of layers in ipyleaflet.

davidbrochart:
But just for my information, is there a way to simply call a JavaScript (blocking) function and get its result in Python?

jasongrout:
See https://ipywidgets.readthedocs.io/en/stable/examples/Widget%20Asynchronous.html for some examples of waiting for user input asynchronously.

PR #914, opened 2022-01-18
implemented open_popup and close_popup methods

PROgram52bc:
#912

### DIFF
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -498,7 +498,7 @@ class Popup(UILayer):
     close_on_escape_key = Bool(True).tag(sync=True, o=True)
 
     def open_popup(self, location=None):
-        """ Open the popup on the bound map
+        """Open the popup on the bound map.
 
         Parameters
         ----------
@@ -511,9 +511,10 @@ class Popup(UILayer):
         self.send({'msg': 'open', 'location': self.location if location is None else location})
 
     def close_popup(self):
-        """ Close the popup on the bound map """
+        """Close the popup on the bound map."""
 
         self.send({'msg': 'close'})
+
 
 class RasterLayer(Layer):
     """Abstract RasterLayer class.
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -497,6 +497,11 @@ class Popup(UILayer):
     auto_close = Bool(True).tag(sync=True, o=True)
     close_on_escape_key = Bool(True).tag(sync=True, o=True)
 
+    def open_popup(self, location=def_loc):
+        self.send({'msg': 'open', 'location': location})
+
+    def close_popup(self):
+        self.send({'msg': 'close'})
 
 class RasterLayer(Layer):
     """Abstract RasterLayer class.
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -497,10 +497,22 @@ class Popup(UILayer):
     auto_close = Bool(True).tag(sync=True, o=True)
     close_on_escape_key = Bool(True).tag(sync=True, o=True)
 
-    def open_popup(self, location=def_loc):
-        self.send({'msg': 'open', 'location': location})
+    def open_popup(self, location=None):
+        """ Open the popup on the bound map
+
+        Parameters
+        ----------
+        location: list, default to the internal location
+            The location to open the popup at.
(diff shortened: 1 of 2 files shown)


ipyleaflet/leaflet.py
```diff
@@ -497,6 +497,11 @@ class Popup(UILayer):
     auto_close = Bool(True).tag(sync=True, o=True)
     close_on_escape_key = Bool(True).tag(sync=True, o=True)
 
+    def open_popup(self, location=def_loc):
+        self.send({'msg': 'open', 'location': location})
```
martinRenou, line 501: There seems to be a duplication of information between the `location` property and the `location` argument of this method.

Should this method default to the location property?
```python
def open_popup(self, location=None):
    self.send({'msg': 'open', 'location': location is None ? self.location : location})
```

And should we change the location property if it's provided?
```python
def open_popup(self, location=None):
    if location is not None:
        self.location = location
    self.send({'msg': 'open', 'location': location is None ? self.location : location})
```

martinRenou: I realize I've been writing non-valid Python code in my comment ahah. Let's try again:

```python
def open_popup(self, location=None):
    self.send({'msg': 'open', 'location': self.location if location is None else location})
```

And should we change the location property if it's provided?
```python
def open_popup(self, location=None):
    if location is not None:
        self.location = location
    self.send({'msg': 'open', 'location': self.location if location is None else location})
```

martinRenou:
flake8 is complaining in the CI test:
```
ipyleaflet/leaflet.py:506:1: E302 expected 2 blank lines, found 1
```

I think we can ignore the UI tests failure, it looks unrelated.

PROgram52bc:
> flake8 is complaining in the CI test:
> 
> ```
> ipyleaflet/leaflet.py:506:1: E302 expected 2 blank lines, found 1
> ```
> 
> I think we can ignore the UI tests failure, it looks unrelated.

Thanks! There is another failure in the Visual Regression test. Do you have any idea what is causing that test to fail?

martinRenou:
> Thanks! There is another failure in the Visual Regression test. Do you have any idea what is causing that test to fail?

It looks like the basemap has changed a bit. You can ignore that. I don't think it's related to your change.

PROgram52bc:
Do you think it is a good idea to also have an `is_open` attribute on the `Popup` class as mentioned in #239? What are some possible ways to implement that?

martinRenou:
You can probably listen to the `popupopen`, and `popupclose` events on the Map https://leafletjs.com/reference-1.3.4.html#map-popupopen

But then you will need to be able to map those events to the right popup widgets?

ipyleaflet/leaflet.py
```diff
@@ -497,6 +497,23 @@ class Popup(UILayer):
...
+        """
+
+        if location is not None:
+            self.location = location
+        self.send({'msg': 'open', 'location': self.location if location is None else location})
+
+    def close_popup(self):
+        """ Close the popup on the bound map """
+
+        self.send({'msg': 'close'})
```
martinRenou, line 517: You need two spaces here for flake8 to be happy

PROgram52bc: Two spaces where?

PROgram52bc:
> But then you will need to be able to map those events to the right popup widgets?

I guess there is a reference to the popup object in the [popup event](https://leafletjs.com/reference.html#popupevent-popup)?

martinRenou:
> I guess there is a reference to the popup object in the popup event?

To the popup object indeed, but there should not be a reference to the popup **widget**

ipyleaflet/leaflet.py
martinRenou: Two blank lines sorry. Between the class definitions.

PROgram52bc:
> > I guess there is a reference to the popup object in the popup event?
> 
> To the popup object indeed, but there should not be a reference to the popup **widget**

I see. I guess we can tackle that issue in a separate PR then.

Any more comments?

PROgram52bc:
> Sorry I wasn't clearer about the flake8 issue. It is still not resolved. I will push a commit to your branch fixing that if you don't mind.

Yeah sure, please feel free :)

martinRenou:
Actually you fixed it! Sorry I was confused by the diff.

Let's wait for the CI to be green then I'll merge :)

Thanks a lot!

PR #945, opened 2022-03-29
Fix Galata screenshot update CI job

martinRenou:


martinRenou:

martinRenou:

PR #955, opened 2022-04-04
Change the theme to pydata-sphynx-theme 

HaudinFlorence:
Change the documentation theme from from sphinx_rtd to pydata-sphynx-theme 
Reorganize the repertories and .rst files to have the proper structure for the documentation


docs/source/layers/index.rst
```diff
@@ -0,0 +1,32 @@
+Layers
+======
+
+.. toctree::
+    :maxdepth: 1
```
martinRenou, line 5:
```suggestion
    :maxdepth: 0
```

So we hopefully don't see the `Examples`, `Usage`, `Attributes` menus in the main page?

martinRenou (changes requested):
It looks way better :D

docs/source/index.rst
```diff
@@ -2,76 +2,12 @@ ipyleaflet: Interactive maps in the Jupyter notebook
 ====================================================
 
 .. toctree::
-    :caption: Installation
     :maxdepth: 2
```
martinRenou, line 5:
```suggestion
    :maxdepth: 1
```

Maybe my other comment is wrong and we should change it here?

martinRenou:
Merging as is, as it's already a great improvement, we can try to tackle my comments later

PR #960, opened 2022-04-08
Add SVGOverlay layer to ipyleaflet

HaudinFlorence:
Add SVGOverlay layer to ipyleaflet


ipyleaflet/leaflet.py
```diff
@@ -744,6 +744,21 @@ class ImageOverlay(RasterLayer):
...
+    Attributes
+    ----------
+    el: string, default ""
+        svg element.
+    """
+
+    _view_name = Unicode('LeafletSVGOverlayView').tag(sync=True)
+    _model_name = Unicode('LeafletSVGOverlayModel').tag(sync=True)
+
+    el = Unicode().tag(sync=True)
```
martinRenou, line 761: Suggestion: `value` instead of `el`? We use `value` in other places for ipywidgets and ipyleaflet.

js/src/layers/SVGOverlay.js
```diff
@@ -0,0 +1,43 @@
...
+
+const DEFAULT_LOCATION = [0.0, 0.0];
+
+export class LeafletSVGOverlayModel extends ImageOverlayModel {
+  defaults() {
+    return {
+      ...super.defaults(),
+      _view_name: 'LeafletSVGOverlayView',
+      _model_name: 'LeafletSVGOverlayModel',
+      svgElement: '',
```
martinRenou, line 15: I guess it should be `el` instead of `svgElement` here? (`value` if you apply my suggestion above)

The comment applies later in the code for `this.model.get('svgElement')`

js/src/layers/SVGOverlay.js
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) Jupyter Development Team.
+// Distributed under the terms of the Modified BSD License.
+
+const L = require('../leaflet.js');
+const rasterlayer = require('./ImageOverlay.js');
+
+const DEFAULT_LOCATION = [0.0, 0.0];
+
+export class LeafletSVGOverlayModel extends ImageOverlayModel {
```
martinRenou, line 9: `ImageOverlayModel` is not defined at this point, I suppose you want to extends `rasterlayer.ImageOverlayModel` instead? Same comment for `ImageOverlayView` later.

js/src/layers/SVGOverlay.js
```diff
@@ -0,0 +1,43 @@
...
+      attribution: ''
+    };
+  }
+}
+
+export class LeafletSVGOverlayView extends ImageOverlayView {
+  create_obj() {
+    this.obj = L.SVGOverlay(
+      this.model.get('svgElement'),
+      this.get_options()
```
martinRenou, line 26: IIUC the second argument for SVGOverlay is the bounds, so maybe:

```javascript
this.obj = L.SVGOverlay(
    this.model.get('el'),
    this.model.get('bounds'),
    this.get_options()
)
```

package.json
```diff
@@ -0,0 +1 @@
+{}
```
martinRenou, line 1: Maybe this has been commited by mistake? Same question for the `countries.geo.json` file

martinRenou (commented):
Thanks! I some comments/questions

davidbrochart:
It looks like you stayed on the same git branch before starting this PR. You should have created a new branch from `master`.
But you can still remove them by rebasing and marking the two first commits as `fixup`, see e.g. https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History.

PR #963, opened 2022-04-13
Fix watch script in ipyleaflet

HaudinFlorence:
In js/package.json :

- Change 

"watch": "webpack --watch"

to
    "watch": "run-p watch:src watch:labextension",
    "watch:src": "webpack --watch",
    "watch:labextension": "jupyter labextension watch ."

- Add in devDependencies : "npm-run-all": "^4.1.5",

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -20,7 +20,9 @@
     "clean": "rimraf dist/ && rimraf ../ipyleaflet/labextension/ && rimraf ../ipyleaflet/nbextension",
     "prepublish": "npm run clean && npm run build",
     "test": "echo \"Error: no test specified\" && exit 1",
-    "watch": "webpack --watch"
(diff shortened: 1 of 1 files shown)


js/package.json
```diff
@@ -20,7 +20,9 @@
     "clean": "rimraf dist/ && rimraf ../ipyleaflet/labextension/ && rimraf ../ipyleaflet/nbextension",
     "prepublish": "npm run clean && npm run build",
     "test": "echo \"Error: no test specified\" && exit 1",
-    "watch": "webpack --watch"
+    "watch": "run-p watch:src watch:labextension",
```
martinRenou, line 23: We are missing the `run-p` command, maybe a missing dev dependency?

martinRenou (changes requested):

PR #964, opened 2022-04-14
Add mention to xyzservices and remove old note

martinRenou:


martinRenou:
Docs are failing because of the new xyzservices which removed the Hydda provider, let's fix this in a separate PR

PR #965, opened 2022-04-15
Simplify max_zoom logics in Map and TileLayer

HaudinFlorence:
Simplify max_zoom logics in Map and TileLayer.

Add changes in Map and TileLayer options to fit the one's of LeafletJS.
Updates of the related docstrings.


js/src/Map.js
```diff
@@ -37,10 +39,10 @@ export class LeafletMapModel extends widgets.DOMWidgetModel {
       _model_module: 'jupyter-leaflet',
       _view_module: 'jupyter-leaflet',
       center: DEFAULT_LOCATION,
```
martinRenou, line 41: You should probably reflect the `center` change you made Python-side by setting a default value of `null` here

js/src/Map.js
```diff
@@ -352,7 +354,30 @@ export class LeafletMapView extends utils.LeafletDOMWidgetView {
...
+                  animate: false
+                });
+                this.dirty = false;
+              }
+              this.model.update_bounds().then(() => {
+                this.touch();
+              });
+            },
+            this
+          );  this.touch();
```
martinRenou, line 380: Can you describe a bit why those changes are needed?

HaudinFlorence:
The changes for the attributes of Map have been made to fit the default values in LeafletJS. According to this zoom and center have now default values put as None. In the notebook, the user has now to give a value for both for the Map to be properly defined.

docs/source/layers/tile_layer.rst
```diff
@@ -33,7 +33,8 @@ Sometimes one could want to specify the date of the given images, for instance w
 
 Attributes and methods
 ----------------------
-
+Note that if you want to display a high resolution layer with a quite large zoom, you have to set ``max_zoom`` and ``max_native_zoom`` with equal value.
```
martinRenou, line 36: Maybe we don't need this and rely only on the docstrings?

ipyleaflet/leaflet.py
```diff
@@ -549,11 +549,13 @@ class TileLayer(RasterLayer):
...
     min_zoom: int, default 0
-        Minimum zoom for this tile service.
+        The minimum zoom level down to which this layer will be displayed (inclusive).
     max_zoom: int, default 18
-        Maximum zoom for this tile service.
-    min_native_zoom: int, default 0
-    max_native_zoom: int, default 18
+        The maximum zoom level up to which this layer will be displayed (inclusive).
+    min_native_zoom: int, default None
+        Minimum zoom number the tile source has available. If it is specified, the tiles on all zoom levels lower than minNativeZoom will be loaded from minNativeZoom level and auto-scaled.
```
martinRenou, line 556:
```suggestion
        Minimum zoom number the tile source has available. If it is specified, the tiles on all zoom levels lower than min_native_zoom will be loaded from min_native_zoom level and auto-scaled.
```

ipyleaflet/leaflet.py
```diff
@@ -549,11 +549,13 @@ class TileLayer(RasterLayer):
...
+    max_native_zoom: int, default None
+        Maximum zoom number the tile source has available. If it is specified, the tiles on all zoom levels higher than maxNativeZoom will be loaded from maxNativeZoom level and auto-scaled.
```
martinRenou, line 558:
```suggestion
        Maximum zoom number the tile source has available. If it is specified, the tiles on all zoom levels higher than max_native_zoom will be loaded from max_native_zoom level and auto-scaled.
```

ipyleaflet/leaflet.py
```diff
@@ -2007,13 +2009,13 @@ class Map(DOMWidget, InteractMixin):
...
         The list of controls that are currently on the map.
-    center: list, default [0, 0]
+    center: list, default None
         The current center of the map.
-    zoom: float, default 12
+    zoom: float, default None
         The current zoom value of the map.
-    max_zoom: int, default 18
+    max_zoom: float, default None
         Maximal zoom value.
```
martinRenou, line 2017:
```suggestion
        Maximum zoom level of the map. If not specified and at least one GridLayer or TileLayer is in the map, the highest of their max_zoom options will be used instead.
```

ipyleaflet/leaflet.py
```diff
@@ -2007,13 +2009,13 @@ class Map(DOMWidget, InteractMixin):
...
-    min_zoom: int, default 1
+    min_zoom: float, default None
         Minimal zoom value.
```
martinRenou, line 2019:
```suggestion
        Minimum zoom level of the map. If not specified and at least one GridLayer or TileLayer is in the map, the lowest of their min_zoom options will be used instead.
```

ipyleaflet/leaflet.py
martinRenou: (taken from the LeafletJS docs)

martinRenou (changes requested):

ipyleaflet/leaflet.py
```diff
@@ -549,16 +549,18 @@ class TileLayer(RasterLayer):
     url: string, default "https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png"
         Url to the tiles service.
     min_zoom: int, default 0
-        Minimum zoom for this tile service.
+        The minimum zoom level down to which this layer will be displayed (inclusive).
     max_zoom: int, default 18
-        Maximum zoom for this tile service.
-    min_native_zoom: int, default 0
-    max_native_zoom: int, default 18
+        The maximum zoom level up to which this layer will be displayed (inclusive). Note that if you want to display a high resolution tile layer with a large zoom, you have to set max_zoom and max_native_zoom with equal value (An example is given in example/Max_zoom_tests.ipynb).
```
martinRenou, line 554:
```suggestion
        The maximum zoom level up to which this layer will be displayed (inclusive).
```

The rest of the comment was not technically true. One can set max_zoom and max_native_zoom with different values and still be able to zoom in on a high resolution tile layer.

Also, pointing to external examples may lead to broken links when renaming/moving example notebooks.

examples/Max_zoom_tests.ipynb
```diff
@@ -0,0 +1,113 @@
+{
```
martinRenou, line 1: Suggestion for the notebook name: `Max_zoom.ipynb`. Also can you clean the cell outputs?

martinRenou (changes requested):

js/src/Map.js
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
 
+import { TileLayer } from 'leaflet';
```
martinRenou, line 4: I suspect this is not needed?

martinRenou (changes requested):

HaudinFlorence:
Please update galata references. Thanks

martinRenou:
Triggering CI

martinRenou:
Thanks!!

PR #966, opened 2022-04-15
Docs remove hydda

HaudinFlorence:
In the documentation and in the examples' notebooks, replace any reference to Hydda by another basemap.


jasongrout:
Thanks! This fixes the doc building error.

PR #969, opened 2022-04-16
Support both processPhosphorMessage and processLuminoMessage

### LINKED ISSUE #773: Issue with processPhosphorMessage method removal
I removed the `processPhosphorMessage` method call and replaced it with `processLuminoMessage`, this was a mistake https://github.com/jupyter-widgets/ipyleaflet/pull/771/files#diff-2262554297993f82eed10102b4e6b1e0157de7d79505a0bc76369905ccd02d78L412. This would have been caught by TypeScript...

We need to patch this and release again.

jasongrout:
This ensures compatibility with ipywidgets 7 and 8.

In https://github.com/jupyter-widgets/ipyleaflet/pull/771, `processPhosphorMessage` was renamed to `processLuminoMessage`, but that renaming should be done at the ipywidgets 7/8 transition, as ipywidgets 7 does not support `processLuminoMessage`.

This PR implements @vidartf's compatibility shim for ipywidgets 7 and 8 that he posted in https://github.com/jupyter-widgets/ipywidgets/pull/3405#pullrequestreview-907246241

Fixes #773

Before this PR, notice the resize does not work in ipywidgets 7.7:

mapbefore

But after this PR, the resize does work in ipywidgets 7.7 (and should still work in ipywidgets 8):

mapafter

### DIFF
--- a/js/src/Map.js
+++ b/js/src/Map.js
@@ -408,8 +408,16 @@ export class LeafletMapView extends utils.LeafletDOMWidgetView {
     }
   }
 
+  processPhosphorMessage(msg) {
+    this._processLuminoMessage(msg, super.processPhosphorMessage);
+  }
+
   processLuminoMessage(msg) {
-    super.processLuminoMessage(msg);
+    this._processLuminoMessage(msg, super.processLuminoMessage);
+  }
+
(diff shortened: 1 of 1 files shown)


martinRenou (approved):
Thanks!

Issue #970, opened 2022-04-17
[Question / Feature] Adding multiple basemaps as leaflet control layers?


whudson:
In leaflet.js it is possible to have multiple basemaps that show up in radio boxes at the top of LayerControl by adding them as control layers:

http://bl.ocks.org/cccruzr/2be38431f91ff2adb9ded7c0e0950382

Is there any way to do this in IPyLeaflet?

whudson:
Passing a dict of TileLayers into the Map() constructor results in no basemap being shown.

whudson:
Adding TileLayers by Map.add_layer() is not ideal because they all load visible on the map.

giswqs:
[Leafmap ](https://github.com/giswqs/leafmap) has some similar functionality. Not sure if it suits your needs. 

Notebook: https://leafmap.org/notebooks/02_using_basemaps
Demo: https://www.youtube.com/watch?v=-lOo-vxjrDM

whudson:
Thanks. There sure are a lot of different options for mapping in Jupyter!

martinRenou:
There is a LayerControl in ipyleaflet, if that's the question: https://ipyleaflet.readthedocs.io/en/latest/controls/layers_control.html

whudson:
That was not my question. IPyLeaflet does not appear to support using multiple basemaps as shown in the linked leaflet.js example.

martinRenou:
IIRC it does, if you add multiple tile layers to the map, giving a name to each tile layer

whudson:
![Capture](https://user-images.githubusercontent.com/519003/164031779-f8eba9b8-763f-4945-b2c9-cbd9c84a7423.PNG)

In the leaflet example. the basemaps are grouped at the top of the later control and selectable using radio buttons. Only one is loaded at a time.

whudson:
Example:
```
from ipyleaflet import Map, basemaps, basemap_to_tiles, LayersControl

m = Map(
    basemap=basemap_to_tiles(basemaps.OpenStreetMap.Mapnik),
    center=(48.204793, 350.121558),
    zoom=3
    )

layer = basemap_to_tiles(basemaps.OpenStreetMap.Mapnik)
layer.name = "Test Layer"
m.add_layer(layer)
m.add_control(LayersControl())
m
```

![Capture](https://user-images.githubusercontent.com/519003/164034845-29227cc9-4cf2-48b3-9de0-fb119edf4973.PNG)

[...]

PR #971, opened 2022-04-20
Document use of multiple basemaps

whudson:
Per the information in https://github.com/jupyter-widgets/ipyleaflet/issues/970 which I have tested and can confirm works, this PR updates the documentation to include an example of using multiple base layers on the TileLayer page, as well as a one sentence reference to it on the Map page.

- [x]  Tested my changes to ensure they built properly in sphinx.

Thank you!


martinRenou:
Thanks!

Issue #428, opened 2019-10-03
Allow nans in choropleth to be mapped to its own unique color and opacity


dwr-psandhu:
branca colormap doesn't deal with nans. A better option would be to allow nans to be mapped to their own unique color and opacity.
See here for [folium option](https://python-visualization.github.io/folium/modules.html#folium.features.Choropleth)

martinRenou:
Thanks for filling this report :)

At some point, I wonder if we'll not drop the branca dependency though. And use bqplot's scales. But dealing with NaNs will still be relevant.

martinRenou:
Should be fixed by https://github.com/jupyter-widgets/ipyleaflet/pull/972

PR #972, opened 2022-04-21
Make it possible to use Choropleth layer with data containing NaNs.

HaudinFlorence:
Make it possible to use Choropleth layer with data containing NaNs. The NaN regions have a dedicated color referred to as nan_color (black is set as default value) and a dedicated opacity referred to as nan_opacity (set to 0.4 as default value).

Resolve issue #428


ipyleaflet/leaflet.py
```diff
@@ -1355,7 +1359,8 @@ def _default_colormap(self):
     def _default_style_callback(self):
         def compute_style(feature, colormap, choro_data):
             return dict(
-                fillColor=colormap(choro_data),
+                fillColor = self.nan_color if math.isnan(choro_data) else colormap(choro_data),
+                fillOpacity = self.nan_opacity if math.isnan(choro_data) else 0.8,
```
martinRenou, line 1363: The linter is complaining about the extra spaces:

```suggestion
                fillColor=self.nan_color if math.isnan(choro_data) else colormap(choro_data),
                fillOpacity=self.nan_opacity if math.isnan(choro_data) else 0.8,
```

martinRenou (changes requested):
Thanks!

martinRenou:
Thanks!!

PR #978, opened 2022-05-02
Add colormap control to be used with Choropleth layer.

HaudinFlorence:
Add a ColormapControl class inherited from WidgetControl class. The widget used is an Output widget.


ipyleaflet/leaflet.py
```diff
@@ -1866,38 +1870,6 @@ def __init__(self, legend, *args, name="Legend", **kwargs):
...
-        self.position = position
-
-    @property
-    def positionning(self):
-        return self.position
-
-    @positionning.setter
-    def positionning(self, position):
-        self.position = position
-
```
martinRenou, line 1900: This part should probably be done in a separate PR?

ipyleaflet/leaflet.py
```diff
@@ -2,16 +2,19 @@
 # Distributed under the terms of the Modified BSD License.
 #
 
+from ast import In
```
martinRenou, line 5: Is this needed?

ipyleaflet/leaflet.py
```diff
@@ -2,16 +2,19 @@
...
 import copy
 import asyncio
 import json
 import xyzservices
 from datetime import date, timedelta
 from math import isnan
+import branca.colormap as cm
+from branca.colormap import linear
```
martinRenou, line 13: This makes ipyleaflet dependent on branca now. So we either need to add it as a dependency in the `setup.py`, or we can find a way to import it lazily for the `ColormapControl` usage.

ipyleaflet/leaflet.py
```diff
@@ -2,16 +2,19 @@
...
 
 from ipywidgets import (
     Widget, DOMWidget, Box, Color, CallbackDispatcher, widget_serialization,
-    interactive, Style
+    interactive, Style, Output, Layout
```
martinRenou, line 17: We probably don't need to import `Layout`?

martinRenou (changes requested):
Thanks!

ipyleaflet/leaflet.py
HaudinFlorence: This was here. I just removed it.

martinRenou: The diff says it's an addition, not a removal

ipyleaflet/leaflet.py
HaudinFlorence: I will add a dependency in the setup.py

ipyleaflet/leaflet.py
HaudinFlorence: Ok, this change has been removed and will be included in a separate PR.

martinRenou:

martinRenou:
Kicking CI

docs/source/controls/colormap_control.rst
```diff
@@ -0,0 +1,64 @@
...
+
+    data_unemployment =  dict(zip(unemployment['State'].tolist(), unemployment['Unemployment'].tolist()))
+
+
+
+    layer = ipyleaflet.Choropleth(
+    geo_data=geo_json_data,
+    choro_data=data_unemployment,
+    colormap=colormap_choice,
+    style={"dashArray": "5, 5"}
```
martinRenou, line 42: Missing indentation

```suggestion
        geo_data=geo_json_data,
        choro_data=data_unemployment,
        colormap=colormap_choice,
        style={"dashArray": "5, 5"}
```

examples/ColormapControl.ipynb
```diff
@@ -0,0 +1,74 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "d822ba7e-2347-4517-a285-159426c71822",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "myColormapControl = ColormapControl(\n",
```
martinRenou, line 40: Same suggestion in the example notebook
```suggestion
    "colormap_control = ColormapControl(\n",
```

PR #979, opened 2022-05-02
Remove get and set decorators in LegendControl in leaflet.py.

HaudinFlorence:
Remove get and set decorators in LegendControl in leaflet.py.


martinRenou:
Thanks! This is a breaking change and we should make the next release a `0.17.0`.

jasongrout:
I added these back in as deprecated things in #1005

PR #985, opened 2022-05-25
Basemap selectors

HaudinFlorence:
Propose in examples 2 notebooks basemaps_with_apikey.ipynb and basemaps_without_apikey.ipynb and a script gallery.py to easily select basemaps among a list and display them. The notebooks deal with the case with and without apikey respectively.

gallery
gallery2


HaudinFlorence:
Would be great to integrate this selector to the documentation but could not find yet a proper way to do it .

PR #988, opened 2022-06-02
Compute the public path automatically

jasongrout:
This updates the public path computation to use built-in require.js facilities, which means we no longer need to hardcode the public path, nor distinguish between the classic notebook extension and the CDN bundle.

See also https://github.com/jupyter-widgets/widget-cookiecutter/pull/103

### DIFF
--- a/js/webpack.config.js
+++ b/js/webpack.config.js
@@ -1,76 +1,63 @@
-var path = require('path');
-var version = require('./package.json').version;
-var crypto = require('crypto');
+var path = require("path");
+var crypto = require("crypto");
 
 // Workaround for loaders using "md4" by default, which is not supported in FIPS-compliant OpenSSL
 var cryptoOrigCreateHash = crypto.createHash;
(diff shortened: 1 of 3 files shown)


js/webpack.config.js
```diff
@@ -1,76 +1,85 @@
...
+    devtool: "source-map",
+    module: {
+      rules: rules,
+    },
+    // 'module' is the magic requirejs dependency used to set the publicPath
+    externals: ["@jupyter-widgets/base", "module"],
+    resolve: resolve,
+  },
+  {
+    // Embeddable {{ cookiecutter.npm_package_name }} bundle
```
martinRenou, line 61:
```suggestion
    // Embeddable jupyter-leaflet bundle
```

martinRenou (changes requested):
Thanks! Just a minor review comment

martinRenou (approved):
Thanks!

Issue #989, opened 2022-06-02
Feature: Add the redraw method for ImageLayer


afonit:
This may be similar to what was implemented for TileLayer:
https://github.com/jupyter-widgets/ipyleaflet/issues/379

I noticed that when I add ImageLayers, then modify the underlying images - those changes don't get reflected in ipyleaflet's ImageLayers.

I imagine if redraw was a method on ImageLayer then when the underlying images are modified I could call redraw on the ImageLayers and they will be updated on the map.

Issue #379, opened 2019-06-11
Add the redraw method for TileLayer


kgiacobbi:
Hi,
It would be nice to add the `redraw()` method for the TileLayer.

In the meantime, i have to use the Cache Busting to force the Layer to re-download the new image and to change the url of the layer.

PR #990, opened 2022-06-02
initial_attempt at resolving redraw for ImageOverlay

afonit:
This is an attempt at resolving:
https://github.com/jupyter-widgets/ipyleaflet/issues/989

I modeled the logic after:
https://github.com/jupyter-widgets/ipyleaflet/issues/379

I was able to follow the steps outlined in:
https://ipyleaflet.readthedocs.io/en/latest/installation/index.html

I have the development environment up and running and the redraw method now appears on ImageOverlay, however when calling redraw that still does not resolve the issue.

I created the pull request to assist in the Issue discussion - if this was not the correct path to develop that conversation please let me know.

### DIFF
--- a/js/src/layers/ImageOverlay.js
+++ b/js/src/layers/ImageOverlay.js
@@ -26,6 +26,7 @@ export class LeafletImageOverlayView extends rasterlayer.LeafletRasterLayerView 
       this.model.get('bounds'),
       this.get_options()
     );
+    this.model.on('msg:custom', this.handle_message.bind(this));
   }
 
   model_events() {
@@ -50,5 +51,11 @@ export class LeafletImageOverlayView extends rasterlayer.LeafletRasterLayerView 
       },
       this
     );
+  }
+
+  handle_message(content) {
+    if (content.msg == 'redraw') {
+      this.obj.redraw();
+    }
   }
 }
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -747,6 +747,8 @@ class ImageOverlay(RasterLayer):
     # Options
     attribution = Unicode().tag(sync=True, o=True)
 
+    def redraw(self):
+        self.send({'msg':'redraw'})
 
 class VideoOverlay(RasterLayer):
     """VideoOverlay class.


js/src/layers/ImageOverlay.js
```diff
@@ -51,4 +52,10 @@ export class LeafletImageOverlayView extends rasterlayer.LeafletRasterLayerView
       this
     );
   }
+
+  handle_message(content) {
+    if (content.msg == 'redraw') {
+      this.obj.redraw();
```
martinRenou, line 58: There doesn't seem to be a `redraw` method on ImageOverlay according to the Leaflet docs https://leafletjs.com/reference.html#imageoverlay-setopacity
unless I missed it?

martinRenou: Maybe you want to use `setUrl` instead, but there apparently no way of forcing the image to update if the url is the same.

According to this stackoverflow comment you can append the date to the url to force an update but I'm not sure we'd want to do this in ipyleaflet's codebase. Though nothing stops you from doing it in your notebook.

martinRenou (changes requested):
Nice thanks!!

js/src/layers/ImageOverlay.js
afonit: Martin, thanks for the review.
For the leafletjs link to imageoverlay - I agree `redraw` is not a method on ImageOverlay, I will search through their github and see if that has been requested/makes sense as a feature request, if not I will file an issue.

The workarounds you suggested are ones I have employed and just seemed to add noise where noise was not needed - so I was trying to implement this redraw.

Since this would first depend on leaflet implementing something - should I just leave this pull request open as I work on that end?  Not sure what the proper protocol is for a pull request.

martinRenou: > The workarounds you suggested are ones I have employed and just seemed to add noise where noise was not needed

Does the workaround work for the `ImageOverlay` then?

> Since this would first depend on leaflet implementing something - should I just leave this pull request open as I work on that end? 

Yes it's fine to keep this PR open for reference. I will turn it into draft.

afonit: > Does the workaround work for the `ImageOverlay` then?
Yes, I was originally just taring down the layer and rebuilding as a workaround.  Your suggestion of just resetting the url is much cleaner - thank you.

It would be nice to not have to get into url management with a new image for each zoom set so I still filed the request on leaflet as that would simplify end-user code.
https://github.com/Leaflet/Leaflet/issues/8277

martinRenou: One thing that would be nice would also to have the ImageOverlay work with an Image (from ipywidgets) widget

PR #991, opened 2022-06-06
Add subitems to a layer (either layers or controls).

HaudinFlorence:
Add subitems to a layer (either layers or controls).


HaudinFlorence:
Please update galata reference

HaudinFlorence:
please update galata reference

Issue #995, opened 2022-06-14
Frontend's _dragging is not recognized by the backend and fails set_state()


xiaochen-db:
Field `_dragging` was added to frontend's `LeafletMapModel` at https://github.com/jupyter-widgets/ipyleaflet/pull/339, however it's not found by the Python backend during [set_state()](https://github.com/jupyter-widgets/ipywidgets/blob/d1491e99f0ed2760f731b0f84dc108721c973fb2/python/ipywidgets/ipywidgets/widgets/widget.py#L570), and thus throws `traitlets.traitlets.TraitError: Class Map does not have a trait named _dragging`. 

Note that this can only be reproduced when env var `JUPYTER_WIDGETS_ECHO` is set to `true`.

Looking at where `_dragging` is used, it seems to be only within the frontend. Can we change it to an attribute rather than the current model field, so that it won't be synced to the backend?

jasongrout:
> Can we change it to an attribute rather than the current model field, so that it won't be synced to the backend?

That was actually [a suggestion](https://github.com/jupyter-widgets/ipyleaflet/pull/339#issuecomment-487632962) on the original ticket that was never implemented. +1 to changing the model `_dragging` value to a view `this._dragging` attribute.

jasongrout:
Note that this bug means that ipyleaflet is broken in ipywidgets 7.7.0 when the widget echo is turned on, and is also broken in master currently. We're fixing ipywidgets to be more lenient, but as @xiaochen-db says, i think it is good to fix it here too.

PR #997, opened 2022-06-15
move _dragging from model value to attribute

xiaochen-db:
This resolves https://github.com/jupyter-widgets/ipyleaflet/issues/995


martinRenou:
Thanks!! This was a leftover from an old implementation that was linking the dragging state in the model

PR #999, opened 2022-06-17
Add Map panes

### LINKED ISSUE #993: Layer order not always respected
In theory, the order of layers on a Map should be determined by the layers python tuple (https://github.com/jupyter-widgets/ipyleaflet/issues/28#issuecomment-367841675, https://github.com/jupyter-widgets/ipyleaflet/issues/843) but in practice this does not always work. Specifically, if I add a GeoJSON layer and a Heatmap layer, the GeoJSON layer is plotted on top of the Heatmap irrespective of the order the layers show up in Map.layers.  Am I doing something wrong?

```python
import os
import json
import requests
from ipyleaflet import Map, GeoJSON, Heatmap

if not os.path.exists('europe_110.geo.json'):
    url = 'https://github.com/jupyter-widgets/ipyleaflet/raw/master/examples/europe_110.geo.json'
    r = requests.get(url)
    with open('europe_110.geo.json', 'w') as f:
        f.write(r.content.decode("utf-8"))

with open('europe_110.geo.json', 'r') as f:
    data = json.load(f)

data['features'] = [data['features'][0]] # Trim to one country so printing layers is readable

m = Map(center=(40.9964, 19.9851), zoom=6)
geo_json = GeoJSON(
    data=data,
    style={
        'opacity': 1, 'dashArray': '9', 'fillOpacity': 0.9, 'weight': 1, 'fillColor': 'white',
    },
)

heatmap = Heatmap(
    locations=[[41.327,19.819],[40.73,19.562]],
    radius=5,
    blur=1,
    min_opacity=1,
    
)
m.add_layer(heatmap)

m.add_layer(geo_json)

m
```
Map looks like this:
Screen Shot 2022-06-07 at 1 57 46 PM

I have the GeoJSON layer in white and mostly opaque so you can see that the red points are behind/below it.

and `[type(x) for x in m.layers]` gives 

[ipyleaflet.leaflet.TileLayer,
 ipyleaflet.leaflet.Heatmap,
 ipyleaflet.leaflet.GeoJSON]

If I reorder the Heatmap and GeoJSON:

`m.layers = [m.layers[0], m.layers[2], m.layers[1]]`

`[type(x) for x in m.layers]` gives

[ipyleaflet.leaflet.TileLayer,
 ipyleaflet.leaflet.GeoJSON,
 ipyleaflet.leaflet.Heatmap]

as expected, but the map looks the same, with the white GeoJSON layer on top of the red points.

davidbrochart:
Fixes #993

### DIFF
--- a/js/src/layers/Layer.js
+++ b/js/src/layers/Layer.js
@@ -23,7 +23,8 @@ export class LeafletLayerModel extends widgets.WidgetModel {
       popup: null,
       popup_min_width: 50,
       popup_max_width: 300,
-      popup_max_height: null
+      popup_max_height: null,
+      pane: ''
     };
   }
 }
@@ -58,6 +59,10 @@ export class LeafletLayerView extends utils.LeafletWidgetView {
       this.listenTo(this.model, 'change:popup', function(model, value) {
         this.bind_popup(value);
       });
+      const pane = this.model.get('pane');
+      if (pane !== '') {
+        L.setOptions(this.obj, {pane});
+      }
     });
   }
 
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -80,6 +80,11 @@ def wait_for_change(widget, value):
     return future
 
 
+class PaneException(TraitError):
+    """Custom PaneException class."""
+    pass
+
+
 class LayerException(TraitError):
     """Custom LayerException class."""
     pass
@@ -112,6 +117,8 @@ class Layer(Widget, InteractMixin):
         Custom name for the layer, which will be used by the LayersControl.
     popup: object
         Interactive widget that will be shown in a Popup when clicking on the layer.
+    pane: string
+        Name of the pane to use for the layer.
     """
 
     _view_name = Unicode('LeafletLayerView').tag(sync=True)
@@ -129,6 +136,7 @@ class Layer(Widget, InteractMixin):
     popup_min_width = Int(50).tag(sync=True)
     popup_max_width = Int(300).tag(sync=True)
     popup_max_height = Int(default_value=None, allow_none=True).tag(sync=True)
+    pane = Unicode('').tag(sync=True)
 
     options = List(trait=Unicode()).tag(sync=True)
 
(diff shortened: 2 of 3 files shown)


js/src/layers/Layer.js
```diff
@@ -111,6 +120,15 @@ export class LeafletLayerView extends utils.LeafletWidgetView {
...
     );
+    this.listenTo(
+      this.model,
+      'change:pane',
+      function() {
+        this.update_pane(),
+        this.map_view.rerender();
+      },
+      this
+    );
```
martinRenou, line 131: We should probably check that this is not triggered at Map creation (so that we don't process two renders when displaying the map the first time)

martinRenou (changes requested):
Looks good!

js/src/layers/Layer.js
davidbrochart: Good point, something like that is happening: when a layer asks the map to re-render (because the layer's pane has changed), I can see that this function is executed more and more. Probably because the previous layers are still hanging around somewhere?

martinRenou: > Probably because the previous layers are still hanging around somewhere?

Oh yeah. We might not have been very careful in ipyleaflet about layers removing. We should probably implement proper `remove` methods in `Views`, which will remove event listeners, delete JS objects etc.

davidbrochart: Fixed in https://github.com/jupyter-widgets/ipyleaflet/pull/999/commits/7293dd54e35e405333b127578b6fb16223d30b63.

davidbrochart: But I'm calling `this.obj.remove()`, so it's on the Leaflet Map, right?

davidbrochart: The whole map is re-created, so the layers too. The previous layers should be garbage-collected, I guess?

davidbrochart: OK to merge?

martinRenou: I see, ok! If there are no side-effects of calling render again then I guess it's good :) 

The readthedocs failure is unrelated

martinRenou (approved):

PR #1001, opened 2022-06-22
Rerender map when changing Map.dragging

### LINKED ISSUE #977: Map dragging attribute cannot be modified after initialization
As mentioned in a [comment on issue #711 ](https://github.com/jupyter-widgets/ipyleaflet/issues/711#issuecomment-704311279), the `dragging` attribute for a map cannot be modified. Specifically, at instance creation, `dragging` can be set to enable/disable dragging, but this cannot be changed later, although the presence of this attribute on the `Map` instance suggests that it should be able to be changed. See the following example:

dragging_bug

davidbrochart:
Fixes #977

It works, but when `dragging` is set the second time, we get this error in the console:
```
272.eb0d340….js:1 Error setting state: Cannot convert undefined or null to object
t.set_state	@	272.eb0d340….js:1
(anonymous)	@	272.eb0d340….js:1
Promise.then (async)		
t._handle_comm_msg	@	272.eb0d340….js:1
_handleCommMsg	@	jlab_core.fc632a0f38…32a0f38747f007842:2
_handleMessage	@	jlab_core.fc632a0f38…32a0f38747f007842:2
await in _handleMessage (async)		
(anonymous)	@	jlab_core.fc632a0f38…32a0f38747f007842:2
Promise.then (async)		
_onWSMessage	@	jlab_core.fc632a0f38…32a0f38747f007842:2
```
Looks like a low-level widget comm message error. Any idea @martinRenou ?

### DIFF
--- a/js/src/Map.js
+++ b/js/src/Map.js
@@ -345,6 +345,19 @@ export class LeafletMapView extends utils.LeafletDOMWidgetView {
     );
     this.listenTo(
       this.model,
+      'change:dragging',
+      function () {
+        if (this.model.get('dragging')) {
+          this.obj.dragging.enable();
+        }
+        else {
+          this.obj.dragging.disable();
+        }
+      },
+      this
+    );
+    this.listenTo(
+      this.model,
       'change:layers',
       function () {
         this.layer_views.update(this.model.get('layers'));


martinRenou:
Do we really need to rerender the entire map just because dragging has been changed? You could probably use `map.dragging.enable();` and `map.dragging.disable();` instead?

davidbrochart:
Indeed!

PR #1002, opened 2022-06-22
Add visible attribute to GeoJSON layer

### LINKED ISSUE #827: Changing GeoJSON/GeoData layer visibility programmatically
Unlike TileLayer, a GeoJSON/GeoData layer does not have the `visible` attribute. Is there a way to turn a GeoJSON/GeoData layer on/off programmatically without using the LayersControl? Thanks.

davidbrochart:
Fixes #827.

### DIFF
--- a/examples/CountriesGeoJSON.ipynb
+++ b/examples/CountriesGeoJSON.ipynb
@@ -50,7 +50,18 @@
    "execution_count": null,
    "metadata": {},
    "outputs": [],
-   "source": []
+   "source": [
+    "layer.visible = False"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "layer.visible = True"
+   ]
   }
  ],
  "metadata": {
@@ -69,7 +80,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.9.7"
+   "version": "3.10.5"
   }
  },
  "nbformat": 4,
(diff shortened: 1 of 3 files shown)


giswqs:
Wow, this is nice! I have been waiting for this feature for a while. Thank you.

davidbrochart:
I've been trying to implement the `visible` attribute using `opacity` and `fillOpacity` before reaching this solution. My conclusion is that it was a mess, because there are global opacities and feature opacities, and once you set the global opacity to 0 you cannot revert.

giswqs:
I would love to contribute. I really need to learn JavaScript. Once this PR is merged and the `visible` attribute is working, I will see if I can make the `opacity` attribute work.

jasongrout:
The docs test is failing for unrelated reasons (404 on a resource)

giswqs:
I just tested v0.17.0 and the GeoJSON `visible` attribute. Just figured out that you can actually control the `style` attribute to set opacity. So it is now possible to change the GeoJSON layer opacity interactively. 

```python
g.style={'fillOpacity': 0.5, 'opacity': 1}
```
image

davidbrochart:
So this is not working since v0.17, right? I guess it was already working before.

giswqs:
@davidbrochart It is working well. I was just saying that the geojson layer opacity can be changed using the `style` attribute, so I don't need to submit a PR to mess up the JS code.

giswqs:
I did not know that `style` attribute before v0.17.0. I was working on the PR and then realized that the `style` attribute is exactly what I needed to change layer opacity interactively, so no PR from me.

davidbrochart:
Great, I just meant that this is not particularly true since v0.17, the `style` attribute was there before.

PR #1004, opened 2022-06-23
Move jupyterlite config to docs directory so it can be found.

jasongrout:
Apparently this config file was just being ignored because it was in the wrong directory.


jasongrout:
The visual regression test failed when rendering a custom leaf icon marker. Perhaps the screenshot was taken too soon? We'll try rerunning the test to see if we pass this time.

martinRenou:
Thanks!

martinRenou (approved):

PR #982, opened 2022-05-16
Change add_layer/remove_layer and add_control/remove_control into add…

HaudinFlorence:
Change add_layer/remove_layer and add_control/remove_control into add/remove working for both layers and controls, in Map Class.

PR #1005, opened 2022-06-23
Add deprecation warnings for API changes

jasongrout:
This is a followup of #979, correcting our LegendControl docs where needed.

I left in a backwards-compatible shim for giving the name argument to the LegendControl, since that is likely the most common thing people would have to change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,13 @@
 * Add Map panes [#999](https://github.com/jupyter-widgets/ipyleaflet/pull/999)
 * Allow setting Map.dragging [#1001](https://github.com/jupyter-widgets/ipyleaflet/pull/1001)
 * Add visible attribute to GeoJSON layer [#1002](https://github.com/jupyter-widgets/ipyleaflet/pull/1002)
-* [BREAKING CHANGE] Remove get and set decorators in LegendControl [#979](https://github.com/jupyter-widgets/ipyleaflet/pull/979)
+* [BREAKING CHANGE] Remove `name`, `legends`, `positioning`, and `positionning` properties in LegendControl [#979](https://github.com/jupyter-widgets/ipyleaflet/pull/979). Update your code with the following substitutions for a LegendControl `legend`:
+  * `legend.name` -> `legend.title`
+  * `legend.legends` -> `legend.legend`
+  * `legend.positioning` -> `legend.position`
+  * `legend.positionnning` -> `legend.position`
+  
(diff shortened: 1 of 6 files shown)


jasongrout:
Also I realized #982 wasn't documented in the changelog, so I did that too.

jasongrout:
I created an 0.17.0 milestone to have the issues I think should be in the release. If those two issues pass review and are merged, I am happy to do the 0.17.0 release:

https://github.com/jupyter-widgets/ipyleaflet/milestone/1

jasongrout:
@HaudinFlorence, you could also review this. Thanks again for the changes cleaning the api up!

ipyleaflet/leaflet.py
```diff
@@ -1932,10 +1932,13 @@ class LegendControl(Control):
         "value 2": "#55A",
         "value 3": "#005"}).tag(sync=True)
 
-    def __init__(self, legend, *args, name="Legend", **kwargs):
+    def __init__(self, legend, *args, **kwargs):
+        kwargs["legend"] = legend
+        # For backwards compatibility with ipyleaflet<=0.16.0
+        if 'name' in kwargs:
+            kwargs.setdefault('title', kwargs['name'])
```
martinRenou, line 1939: Should we show a warning in this case?

jasongrout: We could, but I'm also happy to support it as an undocumented option for now for backwards compatibility. We might ask ourselves why we don't make the other changes just deprecations at this point too, and remove them at some later point, to ease backwards compatibility.

Do you want it deprecated?

jasongrout: I added a bunch of deprecation warnings, and added back the removed apis to be deprecated. I also changed the pendingdeprecationwarnings to just deprecation warnings.

CHANGELOG.md
```diff
@@ -1,13 +1,32 @@
...
 ### New Features
 
 * Make it possible to use Choropleth layer with data containing NaNs [#972](https://github.com/jupyter-widgets/ipyleaflet/pull/972)
 * Add Map panes [#999](https://github.com/jupyter-widgets/ipyleaflet/pull/999)
 * Allow setting Map.dragging [#1001](https://github.com/jupyter-widgets/ipyleaflet/pull/1001)
 * Add visible attribute to GeoJSON layer [#1002](https://github.com/jupyter-widgets/ipyleaflet/pull/1002)
-* [BREAKING CHANGE] Remove get and set decorators in LegendControl [#979](https://github.com/jupyter-widgets/ipyleaflet/pull/979)
 
-## Maintenance
+### Breaking Changes
```
jasongrout, line 12:
```suggestion
### Deprecated API
```

martinRenou (approved):
Thanks!

PR #1006, opened 2022-06-23
Update JupyterLite to include lots of examples

jasongrout:
This greatly expands our JupyterLite examples to include all the examples I saw that were working in JupyterLite.

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -18,5 +18,7 @@ dependencies:
 
   - pip:
     - ..
-    - jupyterlite-sphinx
+    # Until jupyterlite-sphinx>0.4.9 is released, install from a commit hash to get the jupyterlite_contents config option
+    # - jupyterlite-sphinx
+    - https://github.com/jupyterlite/jupyterlite-sphinx/archive/944ccf4ee977627dbf75b5ae1994c08718566c76.zip
     - jupyterlite>=0.1.0b9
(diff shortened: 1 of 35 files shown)


jasongrout:
@martinRenou - I think this is ready for review. I really like all these examples!

Some of the examples I think need to be updated: often they use map.add instead of map.add_control, map.add_layer, etc.

jasongrout:
> Some of the examples I think need to be updated: often they use map.add instead of map.add_control, map.add_layer, etc.

Ah, this is a breaking change in 0.17, documented at https://github.com/jupyter-widgets/ipyleaflet/pull/1005 now. So all is good with these examples, and they will be fixed when we release 0.17.0 and build the docs with that.

martinRenou:
Thanks!

We should probably give xeus-python a shot in this deployment so we don't need the piplite install commands, let's do that in a separate PR

jasongrout:
I'm trying xeus-python for ipywidgets, but I'm running into issues:

1. The lab extensions don't seem to be installed, so widgets are rendering. This is probably something simple to fix, but the bigger point is...
2. The xeus-python download for ipywidgets is something like 40-50MB (and for some reason doesn't seem to be cached for me either). I think that's too heavy (even if cached), and would rather opt for the one extra code cell to get a much lighter-weight jupyterlite.

PR #1007, opened 2022-06-30
Lazy-load the JupyterLite embedded page in documentation

HaudinFlorence:
Remove the retrolite directive in the documentation, since there are already the links and the buttons to click and make an online try.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -1,5 +1,6 @@
 Introduction to ipyleaflet
 ==========================
+
 Ipyleaflet is a `Jupyter widget <https://ipywidgets.readthedocs.io>`_ for
 `Leaflet.js <https://leafletjs.com/>`_ , enabling interactive maps in the
 Jupyter notebook. Every object in ipyleaflet (including the Map, TileLayers, Layers,
@@ -11,10 +12,16 @@ Python or from the browser.
 
 Try it online
 -------------
+
 .. image:: https://jupyterlite.rtfd.io/en/latest/_static/badge.svg
    :target: ./lite/lab?path=ipyleaflet.ipynb
 
 You can try ipyleaflet below, or open many other live examples in a new browser tab with : `JupyterLite <./lite/lab?path=ipyleaflet.ipynb>`_ or `RetroLite <./lite/retro/tree>`_.
+
+.. retrolite:: ipyleaflet.ipynb
+   :width: 100%
+   :height: 810px
+   :prompt: Try Retrolite!
 
 
 Table of Contents
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -21,7 +21,7 @@ You can try ipyleaflet below, or open many other live examples in a new browser 
 .. retrolite:: ipyleaflet.ipynb
    :width: 100%
    :height: 810px
-   :prompt: Try Retrolite!
+   :prompt: Try ipyleaflet!
 
 
 Table of Contents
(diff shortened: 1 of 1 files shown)


jasongrout:
Why would you like to remove the embedded example? For example, is it to simplify the page visually? To make it easier to download?

I think there is a far higher engagement if we embed an example directly in the page, rather than making people click a link. I'm not sure how we'd be able to track that, but I think it's worth embedding an example *and* providing a link to many more examples. Thoughts?

HaudinFlorence:
Thanks for your reply. The main purpose of proposing this change was to make the download easier when a user only wants to have a look at the doc. Also to simplify the page visually (here the index is pushed at the bottom of the page and not really visible anymore). Perhaps we can think about embedding examples another place and redistribute things differently ?

docs/source/index.rst
```diff
@@ -15,10 +16,13 @@ Try it online
...
    :target: ./lite/lab?path=ipyleaflet.ipynb
 
-You can try ipyleaflet below, or open many other live examples in a new browser tab with
-`JupyterLite <./lite/lab?path=ipyleaflet.ipynb>`_ or `RetroLite <./lite/retro/tree>`_.
+You can try ipyleaflet below, or open many other live examples in a new browser tab with : `JupyterLite <./lite/lab?path=ipyleaflet.ipynb>`_ or `RetroLite <./lite/retro/tree>`_.
 
 .. retrolite:: ipyleaflet.ipynb
+   :width: 100%
+   :height: 810px
+   :prompt: Try Retrolite!
```
martinRenou, line 24:
```suggestion
   :prompt: Try ipyleaflet!
```

martinRenou (commented):
Thanks! I just have a minor suggestion :)

docs/source/index.rst
HaudinFlorence: Thanks for the suggestion. This has been changed.

martinRenou:
Thank you!

jasongrout:
> The main purpose of proposing this change was to make the download easier when a user only wants to have a look at the doc. 

My main purpose in reintroducing the example on the front page is to make it much more likely that a user will interact with it. The thought is that every click you require a user to take really cuts down on engagement, and if you make it *so* easy to try things, the user actually will.

I would be very interested in data that could validate that assumption, i.e., how many users will run code in an embedded retrolite, vs how many users will click through to a separate page. My guess is that far more users will run code in an embedded retrolite than will click through to a separate page to run code. However, I think other reasons prevent us from collecting such data.

> Also to simplify the page visually (here the index is pushed at the bottom of the page and not really visible anymore).

What do you think of putting the embedded retrolite behind an HTML `details` element, so that it doesn't take up the vertical space, but still is immediately available to users? I think that takes care of the visual point you mention. Something like:

<details><summary>Try ipyleaflet now</summary>
[The embedded retrolite iframe]
</details>

jasongrout:
It looks like the HTML `details` element has an [event](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#events) for when it is opened, so we could just have a details element that, the first time it is opened, creates the necessary iframe.

jasongrout:
Thanks for introducing the prompt option in this PR. I didn't know about that option, which does seem to solve at least one of the concerns about a heavy page. Awesome!

HaudinFlorence:
> Thanks for introducing the prompt option in this PR. I didn't know about that option, which does seem to solve at least one of the concerns about a heavy page. Awesome!

This was a suggestion of @martinRenou . I did not know about it neither

PR #1010, opened 2022-07-01
Add a colormap control to be used with Choropleth layer

HaudinFlorence:
Add a colormap control to be used with Choropleth layer to display a colormap (the ColormapControl class has Widget Control as a parent and used an Input widget.)


davidbrochart:
Would it be possible to get rid of the right padding?

image

With JupyterLab's dark theme, we can barely see labels. Is it possible to change the label color according to the theme, or to keep a white background and black labels as in the light theme?

image

HaudinFlorence:
> 

I will try to remove the padding. An alternative way to avoid troubles with the dark theme is to set a transparent background (it is an attribute of the WidgetControl).

HaudinFlorence:
Note that when using the current version of Branca, the colormap is rendering as shown above in the comments. Only the min and max values are displayed. We have proposed some changes to customize on Branca colormap, to look like this:
colormap_control

davidbrochart:
Much better, thanks!
I guess dark base maps are still an issue, but at least one now has a bit more control over the background.

image

ipyleaflet/leaflet.py
```diff
@@ -2031,6 +2032,38 @@ def remove_legend_element(self, key):
...
+class ColormapControl(WidgetControl):
+    """ColormapControl class, with WidgetControl as parent class.
+
+    A control which contains a colormap, to be used with Choropleth.
+
+    Attributes
+    ----------
+    caption : str, default 'caption'
+        The caption of the colormap.
+    colormap_choice : str, default 'linear.YlOrRd_04'
```
davidbrochart, line 2044: In the `Choroplet` class, this attribute is called `colormap`, which I prefer.
Also, the type is not a string but the Branca object itself.

davidbrochart (changes requested):

davidbrochart:
I guess you didn't mean to commit `examples/wind-global.nc`?

ipyleaflet/leaflet.py
```diff
@@ -2031,6 +2032,38 @@ def remove_legend_element(self, key):
         self.send_state()
 
 
+class ColormapControl(WidgetControl):
+    """ColormapControl class, with WidgetControl as parent class.
+
+    A control which contains a colormap, to be used with Choropleth.
```
davidbrochart, line 2038: Is it really always meant to be used with a Choroplet layer?

HaudinFlorence:
> Much better, thanks! I guess dark base maps are still an issue, but at least one now has a bit more control over the background.
> 
> image

Yes that's not very visible on the dark map. In that case, I would suggest to set transparent_bg to False.

ipyleaflet/leaflet.py
HaudinFlorence: Not restrictively. I can be extended to any other layers requiring a colormap like for instance the velocity layer. This will be added once the ColormapControl is adopted.

HaudinFlorence:
> I guess you didn't mean to commit `examples/wind-global.nc`?

Indeed. This has been fixed.

ipyleaflet/leaflet.py
HaudinFlorence: The doc string has been changed. This control can be extended to other layers possibly requiring a colormap. Once the control is adopted, it can for instance be applied to a velocity layer.

ipyleaflet/leaflet.py
```diff
@@ -2035,29 +2035,29 @@ def remove_legend_element(self, key):
...
-    A control which contains a colormap, to be used with Choropleth.
+    A control which contains a colormap.
 
     Attributes
     ----------
     caption : str, default 'caption'
         The caption of the colormap.
-    colormap_choice : str, default 'linear.YlOrRd_04'
-        The choosen colormap.
+    colormap: branca.colormap.ColorMap instance, default None
```
davidbrochart, line 2044: Looks like the default is `linear.YlOrRd_04`?

ipyleaflet/leaflet.py
```diff
@@ -2035,29 +2035,29 @@ def remove_legend_element(self, key):
...
+        The colormap used for the effect.
     value_min : float, default 0.0
         The minimal value taken by the data to be represented by the colormap.
     value_max : float, default 1.0
         The maximal value taken by the data to be represented by the colormap.
     """
     caption = Unicode('caption')
-    colormap_choice = Any(linear.YlOrRd_04)
+    colormap = Any(linear.YlOrRd_04)
```
davidbrochart, line 2052: Could be interesting to use traitlets' [Instance](https://traitlets.readthedocs.io/en/stable/trait_types.html#traitlets.Instance)?

HaudinFlorence: Thanks for the comment. Traitlet type and default value in the doc string have been modified in commit

ipyleaflet/leaflet.py
```diff
@@ -5,14 +5,17 @@
 import copy
 import asyncio
 import json
+from folium import LinearColormap
```
davidbrochart, line 8: Should it be imported from `branca`?

ipyleaflet/leaflet.py
```diff
@@ -2031,6 +2033,38 @@ def remove_legend_element(self, key):
...
+        The caption of the colormap.
+    colormap: branca.colormap.ColorMap instance, default linear.YlOrRd_04
+        The colormap used for the effect.
+    value_min : float, default 0.0
+        The minimal value taken by the data to be represented by the colormap.
+    value_max : float, default 1.0
+        The maximal value taken by the data to be represented by the colormap.
+    """
+    caption = Unicode('caption')
+    colormap = Instance(LinearColormap, default_value=linear.YlOrRd_04)
```
davidbrochart, line 2053: Do we require a linear color map, or just any color map?

HaudinFlorence:
please update galata references

ipyleaflet/leaflet.py
HaudinFlorence: We can use Colormap Instance that is more general.

ipyleaflet/leaflet.py
HaudinFlorence: This importation is a mistake (probably coming from the auto completion of VS code)

examples/ColormapControl.ipynb
```diff
@@ -0,0 +1,110 @@
...
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import ipyleaflet\n",
+    "import json\n",
+    "import pandas as pd\n",
+    "from ipywidgets import link, FloatSlider\n",
+    "from branca.colormap import linear\n",
+    "import random\n",
```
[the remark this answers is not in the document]
HaudinFlorence: Panda is used in 'unemployment = pd.read_csv("US_Unemployment_Oct2012.csv")' and the branca importation is also needed here since here we redefined a linear colormap (the same as the default one).
The other unused importation will be removed.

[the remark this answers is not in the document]
davidbrochart: Usually when someone adds a comment it is related to the line just above (unless they refer to several lines).
Here I was mentioning `import random` only.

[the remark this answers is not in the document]
HaudinFlorence: Ok sorry for misunderstanding the comment.

ipyleaflet/leaflet.py
```diff
@@ -2031,6 +2032,38 @@ def remove_legend_element(self, key):
...
+        The caption of the colormap.
+    colormap: branca.colormap.ColorMap instance, default linear.YlOrRd_04
+        The colormap used for the effect.
+    value_min : float, default 0.0
+        The minimal value taken by the data to be represented by the colormap.
+    value_max : float, default 1.0
+        The maximal value taken by the data to be represented by the colormap.
+    """
+    caption = Unicode('caption')
+    colormap = Instance(ColorMap, default_value=linear.YlOrRd_04)
```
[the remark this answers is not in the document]
HaudinFlorence: Thanks for pointing this difference. This has been homogeneized between Choropleth and ColormapControl with commit 4f7a7a9

ipyleaflet/leaflet.py
```diff
@@ -1415,7 +1416,7 @@ class Choropleth(GeoJSON):
     choro_data = Dict()
     value_min = CFloat(None, allow_none=True)
     value_max = CFloat(None, allow_none=True)
-    colormap = Any()
+    colormap = Instance(ColorMap)
```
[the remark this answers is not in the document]
HaudinFlorence: This has been taken into account in commit 4f7a7a9

davidbrochart (approved):
Thanks @HaudinFlorence, that looks good :rocket: 
There seems to be an issue when installing the browser [in the CI](https://github.com/jupyter-widgets/ipyleaflet/runs/7249794821?check_suite_focus=true), but I ran the UI tests locally and they all passed :+1:

HaudinFlorence:
Thanks fa lot or the review @davidbrochart.

PR #1011, opened 2022-07-01
Add subitems to layers (either layers or controls or a mixture of both)

HaudinFlorence:
Add subitems to layers (either layers or controls or a mixture of both).


examples/Choropleth.ipynb
```diff
@@ -65,7 +65,7 @@
     {
      "data": {
       "application/vnd.jupyter.widget-view+json": {
-       "model_id": "6228f855244746258a6ad0553d1326f4",
+       "model_id": "1460d0f40ec14b358ee64285332156f3",
```
davidbrochart, line 68: Maybe not worth committing this file?

examples/Subitems.ipynb
```diff
@@ -0,0 +1,67 @@
...
+    "zoom = 4\n",
+    "geo_json_data = json.load(open(\"us-states.json\"))\n",
+    "m = ipyleaflet.Map(center=center, zoom=zoom)\n",
+    "unemployment = pd.read_csv(\"US_Unemployment_Oct2012.csv\")\n",
+    "unemployment = dict(\n",
+    "    zip(unemployment[\"State\"].tolist(), unemployment[\"Unemployment\"].tolist())\n",
+    ")\n",
+    "\n",
+    "legend = ipyleaflet.LegendControl(\n",
+    "    {\"low\": \"yellow\", \"medium\": \"orange\", \"high\": \"red\"},\n",
```
davidbrochart, line 26: I don't understand, this legend doesn't correspond to the colormap which is used in the choropleth layer (`linear.YlOrRd_04`)?

ipyleaflet/leaflet.py
```diff
@@ -139,6 +139,20 @@ class Layer(Widget, InteractMixin):
...
+
+    @validate('subitems')
+    def _validate_subitems(self, proposal):
+        '''Validate subitems list.
+
+        Makes sure only one instance of any given subitem can exist in the
+        subitem list.
+        '''
+        self._subitem_ids = [subitem.model_id for subitem in proposal.value]
+        if len(set(self._subitem_ids)) != len(self._subitem_ids):
```
davidbrochart, line 153: Do we need `_subitem_ids` to be a class attribute?
```suggestion
        subitem_ids = [subitem.model_id for subitem in proposal.value]
        if len(set(subitem_ids)) != len(subitem_ids):
```

ipyleaflet/leaflet.py
```diff
@@ -1420,6 +1434,7 @@ class Choropleth(GeoJSON):
     nan_color = Unicode('black')
     nan_opacity = CFloat(0.4)
     default_opacity = CFloat(1.0)
+    caption = Unicode('data')
```
davidbrochart, line 1437: Doesn't seem to be used?

davidbrochart (commented):
Could you rebase on master?

HaudinFlorence:
> Could you rebase on master?

Thanks for the review. Rebase is made in

js/src/Map.js
```diff
@@ -208,15 +208,42 @@ export class LeafletMapView extends utils.LeafletDOMWidgetView {
...
+    if(child_view['name'].includes('Control')){
+      this.obj.removeControl(child_view.obj);
+    } else {
+      this.obj.removeLayer(child_view.obj);
+    }
+    child_view.remove();
+    }
+
+    add_subitem_model(child_model) {
+    return this.create_child_view(child_model, {
```
martinRenou, line 230: Small indentation issue

js/src/Map.js
```diff
@@ -244,6 +277,14 @@ export class LeafletMapView extends utils.LeafletDOMWidgetView {
       this.create_panes();
       this.layer_views.update(this.model.get('layers'));
       this.control_views.update(this.model.get('controls'));
+
+      var layer_list = this.model.get('layers');
+      for (let i = 1; i < layer_list.length; i++) { // starting index is 1, since we are not taking into account the TileLayer at index 0
```
martinRenou, line 282: Why are we making a special case for the first layer? It's probably not needed. 

And this line can be re-written `layer_list.forEach((layer) => {`

js/src/Map.js
```diff
@@ -244,6 +277,14 @@ export class LeafletMapView extends utils.LeafletDOMWidgetView {
...
+        var subitem_list = layer_list[i].attributes.subitems
+        this.subitem_views.update(subitem_list);
```
martinRenou, line 284: It looks like this doesn't work if multiple layers have subitems? Only the latest added layer will add its subitems?

You probably need to call `this.subitem_views.update` with the concatenation of all subitems of all layers.

ipyleaflet/leaflet.py
martinRenou: It looks like this was marked resolved by mistake? It looks like it's not resolved

js/src/Map.js
HaudinFlorence: Normally this should have been fixed.

js/src/Map.js
HaudinFlorence: The first layer if not mistaken corresponds to the default basemap in a Map (OpenStreetMap.Mapnik). Initially, we thought we would not add subitems to this layer but you are right it is not necessary or needed to fulfill this condition. We tried to replace the for loop by the forEach proposition.

js/src/Map.js
HaudinFlorence: We tried to implement what is proposed in this comments. When adding 2 layers having their own subitems, they all displayed on the same map. Nevertheless there remains an issue with the method remove_view. `Map.js:221 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'includes')
    at LeafletMapView.remove_subitem_view (Map.js:221:1) `

martinRenou (changes requested):

js/src/Map.js
```diff
@@ -208,15 +208,42 @@ export class LeafletMapView extends utils.LeafletDOMWidgetView {
...
+      this.obj.addLayer(view.obj);
+    }
+
+    // Trigger the displayed event of the child view.
+    this.displayed.then(() => {
+    view.trigger('displayed', this);
+    });
+    return view;
+    });
+    }
```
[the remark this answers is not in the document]
HaudinFlorence: This should be fixed in commit [36d6e74](https://github.com/jupyter-widgets/ipyleaflet/pull/1011/commits/36d6e74eac23f3a918c31c483c513118896699d1)

js/src/Map.js
```diff
@@ -208,15 +208,42 @@ export class LeafletMapView extends utils.LeafletDOMWidgetView {
...
       this.displayed.then(() => {
         view.trigger('displayed', this);
       });
-
       return view;
     });
   }
 
+  remove_subitem_view(child_view) {
+    if(child_view['name'].includes('Control')){
```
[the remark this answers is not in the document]
HaudinFlorence: I agree with the fact that this way of proceeding is a bit week. Since for the moment, in ipyleaflet all control includes the 'Control' subname, this should work but we should indeed think of a better way. Maybe have uniques Add/Remove methods working both for layers and controls ?

js/src/Map.js
```diff
@@ -244,6 +277,20 @@ export class LeafletMapView extends utils.LeafletDOMWidgetView {
...
+      var layer_list = this.model.get('layers');
+      layer_list.forEach((layer) => { // starting index is 1, since we are not taking into account the TileLayer at index 0
+        var all_subitems = []
+        var subitem_list = layer.attributes.subitems
+        subitem_list.forEach((subitem) => {
+          console.log('subitem is:', subitem)
+          all_subitems.push(subitem)
+        }
+        )
+        this.subitem_views.update(all_subitems);
```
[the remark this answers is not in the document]
HaudinFlorence: Thanks for pointing at the mistake of wrong initialization and for the precisions about `update`implementation. This should be fixed in [36d6e74](https://github.com/jupyter-widgets/ipyleaflet/pull/1011/commits/36d6e74eac23f3a918c31c483c513118896699d1)

js/src/Map.js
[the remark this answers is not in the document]
HaudinFlorence: Finally we have updated `add_subitem_model` and `remove_subitem_view` using `instanceof`checks. This is in done in [e514e31](https://github.com/jupyter-widgets/ipyleaflet/pull/1011/commits/e514e31cd34dfb71f55fda8a46bbdab2b110ae60)

js/src/Map.js
```diff
@@ -244,6 +278,17 @@ export class LeafletMapView extends utils.LeafletDOMWidgetView {
...
       this.layer_views.update(this.model.get('layers'));
       this.control_views.update(this.model.get('controls'));
+      var layer_list = this.model.get('layers');
+      var all_subitems = [];
+      layer_list.forEach((layer) => {
+        var subitem_list = layer.attributes.subitems;
+
+        subitem_list.forEach((subitem) => {
+          all_subitems.push(subitem);
+        });
```
martinRenou, line 288: This is equivalent to:

```suggestion
        all_subitems = all_subitems.concat(subitem_list)
```

js/src/Map.js
```diff
@@ -244,6 +278,17 @@ export class LeafletMapView extends utils.LeafletDOMWidgetView {
...
+        this.subitem_views.update(all_subitems);
+      });
```
martinRenou, line 290: This needs to go out of the loop, otherwise you will call `update` as many times as there are layers:

```suggestion
      });
      this.subitem_views.update(all_subitems);
```

ipyleaflet/leaflet.py
```diff
@@ -140,6 +140,20 @@ class Layer(Widget, InteractMixin):
     pane = Unicode('').tag(sync=True)
 
     options = List(trait=Unicode()).tag(sync=True)
+    subitems = Tuple().tag(trait=Instance(Widget), sync=True, **widget_serialization)
+    _subitem_ids = List()
```
martinRenou, line 144: Small leftover from the old implementation:

```suggestion
```

martinRenou (changes requested):
It would be great to add a galata test for this :)

HaudinFlorence:
> It would be great to add a galata test for this :)

A notebook has been added in ui-tests/notebooks in [e0b19ac](https://github.com/jupyter-widgets/ipyleaflet/pull/1011/commits/e0b19ac986774d658b5fb0402ff6f4265dfef6e8), to test adding subitems in layers and check that they correctly display.

HaudinFlorence:
please update galata reference

HaudinFlorence:
Please update galata reference

HaudinFlorence:
please update galata reference

HaudinFlorence:
please update galata references

HaudinFlorence:
The CI tests are now green. Is there other things I may add o check ?

martinRenou:
Thanks! I will give the PR a try. I suspect there is a bug when you create the map in a first Notebook cell, and add the layers with sub-layers in a separate cell.

setup.py
```diff
@@ -58,7 +58,8 @@
         'ipywidgets>=7.6.0,<9',
         'traittypes>=0.2.1,<3',
         'xyzservices>=2021.8.1',
-        'branca>=0.5.0'
+        'branca>=0.5.0',
+        'pandas>=1.4.4'
```
martinRenou, line 62: We don't need pandas as a dependency

```suggestion
        'branca>=0.5.0'
```

martinRenou (changes requested):

setup.py
martinRenou: If that was for fixing the ui tests (that needs pandas), you should instead install pandas in test environment.

martinRenou:
> I suspect there is a bug when you create the map in a first Notebook cell, and add the layers with sub-layers in a separate cell.

Indeed there is an issue when the Map is shown **before** adding the layers with sub-items, as seen in this screenshot (the first map view does not have the subitems visible, the second map view has them):

Screenshot from 2022-09-15 11-48-07

ipyleaflet/leaflet.py
```diff
@@ -2291,6 +2304,8 @@ def _default_options(self):
 
     panes = Dict().tag(sync=True)
     layers = Tuple().tag(trait=Instance(Layer), sync=True, **widget_serialization)
+    controls = Tuple().tag(trait=Instance(Control), sync=True, **widget_serialization)
+
```
martinRenou, line 2308: This is already defined in the class

```suggestion
```

js/src/layers/Layer.js
```diff
@@ -60,6 +91,12 @@ export class LeafletLayerView extends utils.LeafletWidgetView {
         this.bind_popup(value);
       });
       this.update_pane();
+      this;
```
martinRenou, line 94: Probably a leftover

```suggestion
```

js/src/layers/Layer.js
```diff
@@ -60,6 +91,12 @@ export class LeafletLayerView extends utils.LeafletWidgetView {
...
+      this.subitem_views = new widgets.ViewList(
+        this.add_subitem_model,
+        this.remove_subitem_view,
+        this
+      );
```
martinRenou, line 99: You need to call `update` on the view list at least once at creation, so that you populate the sub-items properly.

js/src/layers/Layer.js
```diff
@@ -128,6 +166,16 @@ export class LeafletLayerView extends utils.LeafletWidgetView {
...
+      'change:subitems',
+      function () {
+        console.log('Here we are')
+        this.subitem_views.update(this.subitems);
+      },
+      this
+    );
   }
 
   remove() {
```
martinRenou, line 181: We'll want to clear out all sub-items when removing the layer.

martinRenou (changes requested):

js/src/layers/Layer.js
HaudinFlorence: Ok thanks. this should have been fix in commit [b7e97f9](https://github.com/jupyter-widgets/ipyleaflet/pull/1011/commits/b7e97f9fac9aa201f29b6be649f1947106db5af1)

HaudinFlorence:
please update galata references

HaudinFlorence:
please update galata references

HaudinFlorence:
please update galata references

martinRenou:
please update galata references

martinRenou:
please update galata references

martinRenou:
Please update galata references, this time it should probably work :D

martinRenou:
update galata references

martinRenou:
update galata references

martinRenou:
Triggering CI

martinRenou (approved):
Thank you!!!

HaudinFlorence:
> Thank you!!!

Thanks a lot for the review and help on this PR !

PR #1023, opened 2022-07-16
Add zoom_offset param to TileLayer

### LINKED ISSUE #31: Add tms and zoomOffset options to TileLayer
I'm trying to use some tiles I generated with custom maps. 

I already implemented a pure leaflet test [Leaflet Map Test](http://mariodamore.brapi.net/leaflet/map.html) and I was really excited to get all this in notebooks!

The biggest problem is that the `ipyleaflet.leaflet.TileLayer` cannot change all the parameter `leaflet.tilelayer` ([here](http://leafletjs.com/reference.html#tilelayer)) would accept. 

In particular, the `tms` and `zoomOffset` are critical to my application , I manages to hack the parameter directly in the js code from [jupyter-leaflet](https://www.npmjs.com/package/jupyter-leaflet) on my machine. 

screen shot 2016-06-06 at 13 32 08

Would be hard to implement this in ipyleaflet? In the js code the parameter are there, but commented out.
I haven't enough knowledge of the jupyter widget mechanism  and javascript to make this change.
screen shot 2016-06-06 at 13 32 28

Adding the `tms=true` paramter to `ipyleaflet.leaflet.TileLayer` inizialization doesn't affect the instantiated object, I guess while the class `TileLayer(RasterLayer)` in [leaflet.py](https://github.com/ellisonbg/ipyleaflet/blob/master/ipyleaflet/leaflet.py) doesn't have this parameter declared and cannot pass it further.

ned2:
Add Leaflet's `zoomOffset` param for `TileLayer` class as `zoom_offset`.

Setting `zoom_offset=-1` is necessary in order to be able to support `tile_size=512`, which is the default tile size for Mapbox tiles. 

Fixes #31

### DIFF
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -577,6 +577,8 @@ class TileLayer(RasterLayer):
         Whether the layer is wrapped around the antimeridian.
     tms: boolean, default False
         If true, inverses Y axis numbering for tiles (turn this on for TMS services).
+    zoom_offset: int, default 0
+        The zoom number used in tile URLs will be offset with this value.
     show_loading: boolean, default False
         Whether to show a spinner when tiles are loading.
     loading: boolean, default False (dynamically updated)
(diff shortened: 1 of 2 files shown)


martinRenou (approved):
Thanks a lot!

PR #1027, opened 2022-07-20
Update Binder link in the Usage section of the docs

jtpio:
Update the Binder link in the Usage section of the docs to be the same as the one in the README:

https://mybinder.org/v2/gh/jupyter-widgets/ipyleaflet/stable?urlpath=lab%2Ftree%2Fexamples

This will also open JupyterLab by default instead of the classic notebook.


jtpio:
cc @martinRenou @davidbrochart for review

PR #1031, opened 2022-07-21
property_name was undocumented

jonarmani:


ipyleaflet/leaflet.py
```diff
@@ -2072,6 +2072,8 @@ class SearchControl(Control):
         The marker used by the control.
     found_style: default {‘fillColor’: ‘#3f0’, ‘color’: ‘#0f0’}
         Style for searched feature when searching in LayerGroup.
+    property_name: default None
```
davidbrochart, line 2075: We usually list the type and the default value. Here the type would be `string`, and it would default to `"display_name"`, not `None`.

davidbrochart (changes requested):

PR #1033, opened 2022-07-31
Add support for `as_leaflet_layer` in `Map.add`

banesullivan:
I use `ipyleaflet` a ton with [localtileserver](https://github.com/banesullivan/localtileserver) and I was thinking it'd be convenient if `ipyleaflet.Map`'s `add()` method supported passing objects with an `as_leaflet_layer()` interface. 

There may be better/cleaner ways to do this, so suggestions are welcome! But, my end goal is to make it easier to go from raster file path to `ipyleaflet.Map` showing tiles like:

```py
from localtileserver import TileClient
from ipyleaflet import Map

client = TileClient('path/to/geo.tif')  # <--- will have `as_leaflet_layer` method

m = Map(center=client.center(), zoom=client.default_zoom)
m.add_layer(client)
m
```

### DIFF
--- a/docs/source/layers/layer_like.rst
+++ b/docs/source/layers/layer_like.rst
@@ -4,7 +4,7 @@ Layer-Like Objects
 :class:`ipyleaflet.Map`'s :func:`ipyleaflet.Map.add` method supports
 "layer-like" objects; meaning any object with an ``as_leaflet_layer`` method.
 This interface can be especially useful for downstream developers who want
-their users to more easily be able to add thier objects to an
+their users to more easily be able to add their objects to an
 :class:`ipyleaflet.Map`.
 
 Example
--- a/docs/source/layers/layer_like.rst
+++ b/docs/source/layers/layer_like.rst
@@ -24,17 +24,9 @@ Here is a simple example of creating a custom data class to hold heatmap data
 
   class MyHeatMap:
       def __init__(self, points, values, radius=20):
-          self._points = points
-          self._values = values
-          self._radius = 20
-
-      @property
-      def points(self):
-          return self._points
-
-      @property
-      def values(self):
-          return self._counts
+          self.points = points
+          self.values = values
+          self.radius = 20
 
       @property
       def data(self):
@@ -44,21 +36,26 @@ Here is a simple example of creating a custom data class to hold heatmap data
           from ipyleaflet import Heatmap
           return Heatmap(
               locations=self.data.tolist(),
-              radius=self._radius,
+              radius=self.radius,
           )
+
+We can now use that custom data class and because it has an
+``as_leaflet_layer`` interface, we can pass the object directly to
+:func:`ipyleaflet.Map.add`.
+
 
 .. jupyter-execute::
 
   from ipyleaflet import Map
 
   n = 1000
-  d = MyHeatMap(
+  data = MyHeatMap(
       np.random.uniform(-80, 80, (n, 2)),
       np.random.uniform(0, 1000, n),
   )
 
   m = Map(center=(0, 0), zoom=2)
-  m.add_layer(d.as_leaflet_layer())
+  m.add(data)
   m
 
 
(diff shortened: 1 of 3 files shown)


davidbrochart (approved):
Thanks for the suggestion @banesullivan, I think it would make sense, including for [xarray-leaflet](https://github.com/xarray-contrib/xarray_leaflet).

ipyleaflet/leaflet.py
```diff
@@ -1170,6 +1170,8 @@ def add(self, layer):
 
         if isinstance(layer, dict):
             layer = basemap_to_tiles(layer)
+        elif hasattr(layer, 'as_leaflet_layer'):
```
martinRenou, line 1173: Could you add a comment saying that it's specific to `localtileserver`?

martinRenou (commented):
Thank you!

ipyleaflet/leaflet.py
davidbrochart: We might want to generalize this interface? I think it could make sense outside `localtileserver` too, for instance in `xarray-leaflet`.

martinRenou: Yeah that makes sense!

davidbrochart:
@banesullivan I think we should document this interface.

ipyleaflet/leaflet.py
banesullivan: My intention is definitely for this to be generally applicable. 

I'd love to see libraries like `shapely` have an `as_leaflet_layer` too (or perhaps I'll just monkey path that method) to make my life easier

banesullivan:
> I think we should document this interface.

Agreed. If you all are happy with `as_leaflet_layer` being the name and it being a method, then I'll add some docs when I get a few minutes

docs/source/layers/layer_like.rst
```diff
@@ -0,0 +1,68 @@
+Layer-Like Objects
```
banesullivan, line 1: @davidbrochart and @martinRenou, please review this explanation and feel free to modify. This is what I came up with on a whim to document the interface.

docs/source/layers/layer_like.rst
```diff
@@ -0,0 +1,68 @@
...
+==================
+
+:class:`ipyleaflet.Map`'s :func:`ipyleaflet.Map.add` method supports
+"layer-like" objects; meaning any object with an ``as_leaflet_layer`` method.
+This interface can be especially useful for downstream developers who want
+their users to more easily be able to add thier objects to an
```
davidbrochart, line 7:
```suggestion
their users to more easily be able to add their objects to an
```

banesullivan: Thanks! Wrote this up rather quickly

docs/source/layers/layer_like.rst
```diff
@@ -0,0 +1,68 @@
...
+.. jupyter-execute::
+
+  import numpy as np
+
+
+  class MyHeatMap:
+      def __init__(self, points, values, radius=20):
+          self.points = points
+          self.values = values
+          self.radius = 20
```
davidbrochart, line 29:
```suggestion
          self.radius = radius
```

docs/source/layers/layer_like.rst
```diff
@@ -0,0 +1,68 @@
...
+  m
+
+
+External Examples
+-----------------
+
+The following external libraries are working to implement this new interface
+
+- `localtileserver <https://github.com/banesullivan/localtileserver>`_: a dynamic tile server built for visualizing large geospatial images/rasters with ipyleaflet.
+- `xarray-leaflet <https://github.com/davidbrochart/xarray_leaflet>`_: an xarray extension for tiled map plotting, based on ipyleaflet.
```
davidbrochart, line 68:
```suggestion
- `xarray-leaflet <https://github.com/xarray-contrib/xarray_leaflet>`_: an xarray extension for tiled map plotting, based on ipyleaflet.
```

davidbrochart:
@banesullivan I just made a small change to the docs, otherwise it looks good to me.
Thanks!

banesullivan:
Thank you for the second set of eyes and improvements!

davidbrochart:
I opened #1037, the code was `LayerGroup` instead of `Map` :smile:

PR #1035, opened 2022-08-04
Adapt velocity layer for an easier use with a ColormapControl

HaudinFlorence:
Adapt velocity layer re-building the color_scale list from a linear colormap of Branca. This enables to easily use a ColormapControl with this layer using the same colormap in the layer and in the control.


ipyleaflet/velocity.py
```diff
@@ -67,20 +67,17 @@ class Velocity(Layer):
...
+
+    @default('color_scale')
+    def _default_color_scale(self):
+        self.color_scale = []
+
+        for i in range(len(self.colormap.colors)):
+            rgb_tuple = tuple(str(int(x * 256)) for x in self.colormap.colors[i][:3])
+            rgb_tuple_str = " , ".join(rgb_tuple)
+            rgb_str = 'rgb(' + rgb_tuple_str + ')'
+            self.color_scale.append(rgb_str)
```
davidbrochart, line 81:
```suggestion
        for color in self.colormap.colors:
            rgb_tuple = tuple(int(x * 256) for x in color[:3])
            rgb_str = f'rgb({" , ".join(rgb_tuple)})'
            self.color_scale.append(rgb_str)
```

ipyleaflet/velocity.py
```diff
@@ -76,7 +76,7 @@ def _default_color_scale(self):
 
         for color in self.colormap.colors:
             rgb_tuple = tuple(int(x * 256) for x in color[:3])
-            rgb_str = f'rgb({" , ".join(rgb_tuple)})'
+            rgb_str = f'rgb{str(rgb_tuple)}'
```
davidbrochart, line 79:
```suggestion
            rgb_str = f"rgb{rgb_tuple}"
```

examples/Velocity_with_colormap.ipynb
```diff
@@ -0,0 +1,157 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Set up for JupyterLite\n",
+    "try:\n",
+    "    import piplite\n",
+    "    await piplite.install(['ipyleaflet')\n",
```
davidbrochart, line 23:
```suggestion
    "    await piplite.install('ipyleaflet')\n",
```

ipyleaflet/velocity.py
```diff
@@ -67,20 +67,16 @@ class Velocity(Layer):
...
+    @default('color_scale')
+    def _default_color_scale(self):
+        self.color_scale = []
+
+        for color in self.colormap.colors:
+            rgb_tuple = tuple(int(x * 256) for x in color[:3])
+            rgb_str = f"rgb{rgb_tuple}"
+            self.color_scale.append(rgb_str)
+
+        return self.color_scale
```
davidbrochart, line 82: By default we don't have the same color scale as before, maybe we should?
Also, it might be good to check that we don't pass a `colormap` *and* a `color_scale`.

davidbrochart:
Thanks @HaudinFlorence!
It could be good to update the documentation also?

HaudinFlorence:
> Thanks @HaudinFlorence! It could be good to update the documentation also?

The docstrings have been updated.

ipyleaflet/velocity.py
HaudinFlorence: Yes, you are right that we don't have the same default `color scale`as before. Here we propose that this colormap is defined from the colormap in such a way that there is a total correspondance between the mapping of colors in the colormap and the mapping of colors on the map. To avoid any confusion, we propose that the user only has to give a colormap and we make the color_scale a private attribute (change proposed in [ed48e77](https://github.com/jupyter-widgets/ipyleaflet/pull/1035/commits/ed48e7785aa650a471cf0f199051c28ed4b260fc) )

examples/Velocity_with_colormap.ipynb
davidbrochart: You reintroduced the `[`.

js/src/layers/Velocity.js
```diff
@@ -25,7 +25,7 @@ export class LeafletVelocityModel extends layer.LeafletLayerModel {
       minVelocity: 0,
       maxVelocity: 10,
       velocityScale: 0.005,
-      colorScale: []
+      _colorScale: []
```
davidbrochart, line 28: Actually the `VelocityLayer` expects [colorScale](https://github.com/onaci/leaflet-velocity/blob/ab97f77c237e60ff7bb1f5bb30335c6e583d9209/src/js/L.VelocityLayer.js#L10), so this won't work, unless you copy the value of `_colorScale` to the `colorScale` option.

HaudinFlorence: We have reverted the change of making color_scale a private attribute.

ipyleaflet/velocity.py
```diff
@@ -38,49 +38,68 @@ class Velocity(Layer):
...
+        colormap_control = ColormapControl(
+            caption=self.caption,
+            colormap=self.colormap,
+            value_min=self.min_velocity,
+            value_max=self.max_velocity,
+            position="topright",
+            transparent_bg=False,
+        )
+        self.subitems = (colormap_control,)
+        return self.subitems
```
martinRenou, line 105: The `@default` method will set `self.subitems` automatically, you should not set it yourself. See https://traitlets.readthedocs.io/en/stable/using_traitlets.html#default-values-and-checking-type-and-value

```suggestion
        return (colormap_control,)
```

ipyleaflet/velocity.py
```diff
@@ -38,49 +38,68 @@ class Velocity(Layer):
...
+    @default("color_scale")
+    def _default_color_scale(self):
+        self.color_scale = []
+
+        for color in self.colormap.colors:
+            rgb_tuple = tuple(int(x * 256) for x in color[:3])
+            rgb_str = f"rgb{rgb_tuple}"
+            self.color_scale.append(rgb_str)
+
+        return self.color_scale
```
martinRenou, line 92: Some suggestions/issues here:
- you should not set `self.color_scale` yourself (same comment as for `subitems`)
- we must multiply by `255` because we scale the range from `[0., 1.]` to `[0, 255]` (256 colors)
- we can use list comprehension here instead of a for-loop
- we can/should probably use `rgba` if that's how the colormap is defined

```suggestion
        return [
            f"rgba{tuple(int(x * 255) for x in color)}"
            for color in self.colormap.colors
        ]
```

ipyleaflet/velocity.py
```diff
@@ -38,49 +38,68 @@ class Velocity(Layer):
...
-        "rgb(252,150,75)",
-        "rgb(250,112,52)",
-        "rgb(245,64,32)",
-        "rgb(237,45,28)",
-        "rgb(220,24,32)",
-        "rgb(180,0,35)"
-    ]).tag(sync=True, o=True)
+    colormap = Any(linear.OrRd_06)
+    color_scale = List([]).tag(sync=True, o=True)
+    caption = Unicode("").tag(sync=True, o=True)
```
martinRenou, line 81: We don't need to sync this one traits with the front-end?

```suggestion
    color_scale = List([]).tag(sync=True, o=True)
    caption = Unicode("").tag(o=True)
```

js/src/layers/Velocity.js
```diff
@@ -26,6 +26,7 @@ export class LeafletVelocityModel extends layer.LeafletLayerModel {
       maxVelocity: 10,
       velocityScale: 0.005,
       colorScale: [],
+      caption:''
```
martinRenou, line 29: If we don't sync caption, it should not be part of the model

```suggestion
```

ipyleaflet/velocity.py
```diff
@@ -38,49 +38,68 @@ class Velocity(Layer):
...
-        "rgb(252,217,125)",
-        "rgb(255,182,100)",
-        "rgb(252,150,75)",
-        "rgb(250,112,52)",
-        "rgb(245,64,32)",
-        "rgb(237,45,28)",
-        "rgb(220,24,32)",
-        "rgb(180,0,35)"
-    ]).tag(sync=True, o=True)
+    colormap = Any(linear.OrRd_06)
```
martinRenou, line 79: It would be nice to be able to change this trait dynamically

```python
velocity_layer.colormap = linear.Accent_04
```

(same comment for caption)

martinRenou (changes requested):

PR #1037, opened 2022-08-10
Fix as_leaflet_layer

davidbrochart:
Follow-up to #1033.

### DIFF
--- a/ipyleaflet/leaflet.py
+++ b/ipyleaflet/leaflet.py
@@ -1172,8 +1172,6 @@ class LayerGroup(Layer):
 
         if isinstance(layer, dict):
             layer = basemap_to_tiles(layer)
-        elif hasattr(layer, 'as_leaflet_layer'):
-            layer = layer.as_leaflet_layer()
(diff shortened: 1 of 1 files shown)


davidbrochart:
There seems to still be an issue in the docs though.

davidbrochart:
My bad, the docs are fine locally. I'm going to revert the changes to the docs so that it's as in #1033 @banesullivan.

ipyleaflet/leaflet.py
```diff
@@ -1172,8 +1172,6 @@ def add(self, layer):
 
         if isinstance(layer, dict):
             layer = basemap_to_tiles(layer)
-        elif hasattr(layer, 'as_leaflet_layer'):
```
martinRenou, line 1175: We should probably keep it here too though.

I wonder if we could implement `remove` and `substitute` too.

davidbrochart: Yes, let's do all that in follow-up PRs.

PR #1039, opened 2022-08-17
Remove HikeBike basemap from the examples

HaudinFlorence:
Remove HikeBike basemap from the examples shown in the documentation, since it seems not to be maintained anymore.


martinRenou:
Thanks!

Issue #1020, opened 2022-07-12
Map doesn't show in JupyterLite


davidbrochart:
When browsing to https://ipyleaflet.readthedocs.io/en/latest:
![image](https://user-images.githubusercontent.com/4711805/178437565-3e952546-80a0-47ad-afa8-6b7e53820dc6.png)

davidbrochart:
It works in https://jupyterlite.readthedocs.io/en/latest/_static/lab, but it has ipyleaflet v0.16:
![image](https://user-images.githubusercontent.com/4711805/178438205-13054d39-e9d3-47b7-985e-15be119bf275.png)

jtpio:
It's probably because the deployed version pulled `0.16` at build time for the frontend extension.

And now calling `piplite.install('ipyleaflet')` pulls `0.17` from PyPI, but the frontend assets are still the same (`0.16`). 

There are multiple ways of solving this. Either by pinning the version with `piplite.install('ipyleaflet==0.17')` in the notebook and making sure to update it on new releases.

Or use an approach like in the `ipywidgets` repo with `piplite_urls`: https://github.com/jupyter-widgets/ipywidgets/blob/master/docs/source/jupyter_lite_config.json

Or switch to the `jupyterlite-xeus-python` kernel and also make sure the version are pinned or at least in sync.

HaudinFlorence:
The first solution is implemented in https://github.com/jupyter-widgets/ipyleaflet/pull/1040.

jasongrout:
> Or use an approach like in the `ipywidgets` repo with `piplite_urls`: https://github.com/jupyter-widgets/ipywidgets/blob/master/docs/source/jupyter_lite_config.json

One really nice thing about this approach is that it works for master as well, in that it bundles exactly the version of ipyleaflet the repo is currently at, so you can immediately test the current state of master.

Note that in addition to the jupyter_lite_config.json part, you also need to build the dev version of the plugin: https://github.com/jupyter-widgets/ipywidgets/blob/58adde2cfbe2f78a8ec6756d38a7c637f5e599f8/docs/source/conf.py#L151-L171

HaudinFlorence:
I would be happy to make this fix with the `piplite_urls`. I have been having a look but I haven't managed to fully transpose the logics shown for ipywidgets repo to the case of ipyleaflet.

PR #1040, opened 2022-08-17
Fix map not displaying in jupyterlite.

HaudinFlorence:
Fix issue https://github.com/jupyter-widgets/ipyleaflet/issues/1020. The first solution proposed by @jtpio in this issue is implemented, that is "pinning the version with piplite.install('ipyleaflet==0.17') in the notebook."

### DIFF
--- a/docs/source/ipyleaflet.ipynb
+++ b/docs/source/ipyleaflet.ipynb
@@ -8,7 +8,7 @@
    "outputs": [],
    "source": [
     "import piplite\n",
-    "await piplite.install('ipyleaflet')"
+    "await piplite.install('ipyleaflet == 0.17')"
    ]
   },
   {


docs/source/ipyleaflet.ipynb
```diff
@@ -8,7 +8,7 @@
    "outputs": [],
    "source": [
     "import piplite\n",
-    "await piplite.install('ipyleaflet')"
+    "await piplite.install('ipyleaflet == 0.17')"
```
jtpio, line 11: Maybe the extra spaces around the `==` could be trimmed?

```suggestion
    "await piplite.install('ipyleaflet==0.17')"
```

jtpio (approved):
Nice thanks @HaudinFlorence for fixing this :+1: 

Just left a minor comment above.

martinRenou:
Another way to fix this could be to use xeus-python :D

docs/source/ipyleaflet.ipynb
HaudinFlorence: Thanks.This has been changed.

HaudinFlorence:
> Another way to fix this could be to use xeus-python :D

I would be like to do it but I don't know how.

PR #1041, opened 2022-08-18
Restrict ipywidgets version in CI

martinRenou:
Just trying to see if I can replicate the UI tests failure from https://github.com/jupyter-widgets/ipyleaflet/pull/968


martinRenou:
So it really seems to be an issue with the ipywidgets 8 release

PR #1044, opened 2022-08-26
Add a browser entry point in package.json

jasongrout:
By default, if we want to use ipyleaflet by itself in the browser (like voila, etc.), we want to serve up the AMD module. This points the default browser import to the AMD module. jsdelivr recognizes this, and we'll see if unpkg recognizes it too.

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -15,6 +15,7 @@
   "license": "MIT",
   "author": "Project Jupyter",
   "main": "src/index.js",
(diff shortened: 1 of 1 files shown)


jasongrout:
Converting to draft since it's not clear the full ramification of this. For example, will webpack start importing the dist/index.js? It may be better to do the more conservative:

```
"unpkg": "dist/index.js",
"jsdelivr": "dist/index.js"
```

PR #1054, opened 2022-09-22
Update the way we run UI-tests

martinRenou:
This uses a logic closer to other projects (ipympl, bqplot, ipydatagrid). And run ui-tests with the dark theme.


martinRenou:

martinRenou:

PR #1132, opened 2023-08-07
subdomains for tile.openstreetmap.org are deprecated

mtmail:
OpenStreetMap operations teams deprecated some tile server URL formats. Removing the subdomains is also better for browser caching. https://github.com/openstreetmap/operations/issues/737


Firefishy:
@martinRenou change look ok?

martinRenou:
Triggering CI

PR #1133, opened 2023-08-12
Sanitize `style` attribute in DrawControl.js for `data` attribute serialization

### LINKED ISSUE #1119: Draw control can't stop drawing after draw more than one shape
Hey, 

I am using ipyleaflet for my work. I would like to draw rectangles on my map with DrawControl. 
```Python 

map = Map(center=(49.115469, -1.082814), zoom=10)
dc = DrawControl(
    marker={"shapeOptions": {"color": "#0000FF"}},
    rectangle={"shapeOptions": {"color": "#0000FF"}},
    circle={"shapeOptions": {"color": "#0000FF"}},
    circlemarker={},
)
map.add(dc)
display(map)
``` 
The first drawing works but when I click again on the rectangle for example, DrawControl won't stop drawing rectangles until I clear them all.

The above code reproduces the bug. 

Here is what I get.
[Screencast from 19-06-2023 18:39:08.webm](https://github.com/jupyter-widgets/ipyleaflet/assets/76174883/3321fa9c-f5b6-4b49-8445-2c91c9678a9b)

sufyanAbbasi:
- Fixes #1119 1119

- The main issue is that a `style()` function is added to the GeoJSON object `style` property in `create_obj()`, which triggers an error when the `data` attribute is serialized here: https://github.com/jupyter-widgets/ipywidgets/blob/e1718c2b3bf0b143580ef87f71c55fbc6ed50a77/packages/base/src/widget.ts#L587

- This is because `structuredClone` throws an error for functions: https://web.dev/structured-clone/#features-and-limitations. While the article mentions drawbacks for using `JSON.parse(JSON.stringify(...))`, I believe that this component does not produce any non-primitive properties, so this is a valid strategy for removing offending lines.

- Tested by placing a breakpoint and testing directly on the browser and then building a local ipyleaflet in Jupyter Lab.

### DIFF
--- a/js/src/controls/DrawControl.js
+++ b/js/src/controls/DrawControl.js
@@ -141,7 +141,10 @@ export class LeafletDrawControlView extends control.LeafletControlView {
     let newData = [];
     this.feature_group.eachLayer((layer) => {
       const geoJson = layer.toGeoJSON();
-      geoJson.properties.style = layer.options;
+      // Sanitize layer options for serialization via `structuredClone`:
+      // https://web.dev/structured-clone/#features-and-limitations
+      const sanitizedLayerOptions = JSON.parse(JSON.stringify(layer.options));
+      geoJson.properties.style = sanitizedLayerOptions;
       newData.push(geoJson);
     });
     this.model.set('data', newData);


nyehughes:
I was still having problems with the draw control, as described here https://github.com/jupyter-widgets/ipyleaflet/issues/1119,  even with this fix applied. (I using ipyleaflet running under the Solora framework).  However, using the same technique proposed in this PR, but applied further down the code, line 144, fixed it for me:  ```
layers_to_data() {
    let newData = [];
    this.feature_group.eachLayer((layer) => {
      const geoJson = layer.toGeoJSON();
      // geoJson.properties.style = layer.options; // REMOVE LINE
      const sanitizedLayerOptions = JSON.parse(JSON.stringify(layer.options)); //ADD
      geoJson.properties.style = sanitizedLayerOptions; //ADD
      newData.push(geoJson);
    });
    this.model.set('data', newData);
    this.model.save_changes();
  }
```

PR #1135, opened 2023-09-07
update leaflet-splitmap to v1.0.3

oxidase:
Fixes: #1066 

https://github.com/jupyter-widgets/ipyleaflet/assets/4421046/54c6bf12-6dab-49af-a960-0c4118238979

@martinRenou fyi

PR #1136, opened 2023-09-08
Use micromamba

oxidase:
Use micromamba as suggested in https://github.com/jupyter-widgets/ipyleaflet/pull/1135#issuecomment-1711265171


oxidase:
@martinRenou should be green now

PR #1139, opened 2023-10-07
Removed the unwanted character from Python code.

frankwwu:
Removed the unwanted character from Python code.


martinRenou (approved):
thks

PR #1150, opened 2023-12-05
Add visibility to MarkerCluster

gjmooney:


js/src/layers/MarkerCluster.js
```diff
@@ -34,6 +34,30 @@ export class LeafletMarkerClusterView extends layer.LeafletLayerView {
       },
       this
     );
+    this.listenTo(
+      this.model,
+      'change:opacity',
+      function () {
+        if (this.model.get('visible')) {
+          document.querySelector('.test-icon').style.opacity =
```
martinRenou, line 42: This would not work if you have multiple marker clusters on the map and want to hide one and not the other. Hiding one would hide the other and that's unfortunately not what you want.

js/src/layers/MarkerCluster.js
```diff
@@ -52,12 +76,35 @@ export class LeafletMarkerClusterView extends layer.LeafletLayerView {
 
   create_obj() {
     var options = this.get_options();
-    this.obj = L.markerClusterGroup(options);
+    this.obj = L.markerClusterGroup({
+      disableClusteringAtZoom: options['disableClusteringAtZoom'],
+      maxClusterRadius: options['maxClusterRadius'],
```
martinRenou, line 81: Instead of passing options manually you can unroll all of them with the following
```suggestion
    this.obj = L.markerClusterGroup({
      ...options,
```

martinRenou (changes requested):
I'm not sure about this approach, this feels a bit hacky.

Another approach could be to loop over all the markers and apply the same opacity on all of them.

PR #1151, opened 2023-12-08
Marker cluster api docs

gjmooney:
Expose more of the MarkerCluster API to ipyleaflet and update the docs with the additional options.


martinRenou (approved):
🚀 Thanks!

PR #1154, opened 2023-12-11
Documentation: Fix CSS issue

gjmooney:
Fix for the bug mentioned in [#1578](https://github.com/pydata/pydata-sphinx-theme/issues/1578)

### DIFF
--- a/docs/_static/main_stylesheet.css
+++ b/docs/_static/main_stylesheet.css
@@ -1,4 +1,8 @@
-.wy-nav-content{
-    max-width: 1000px;
-    margin: auto;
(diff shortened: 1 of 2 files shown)


martinRenou:
Thanks! It seems to have fixed properly the Image Overlay docs

image

There is still an issue with the icon though: https://ipyleaflet--1154.org.readthedocs.build/en/1154/layers/icon.html (there should not be a background on the marker)

gjmooney:
The new changes should make markers and icons display correctly as well.

martinRenou (approved):
Nice! Thank you

PR #1156, opened 2023-12-14
Add position to Maps add method

gjmooney:
Add ``position`` parameter to Maps ``add`` method, allowing users to specify where in the layers stack they would add their new layer.


ipyleaflet/leaflet.py
```diff
@@ -2676,13 +2676,15 @@ def __isub__(self, item):
     def __add__(self, item):
         return self.add(item)
 
-    def add(self, item):
+    def add(self, item, position=None):
```
martinRenou, line 2679:
```suggestion
    def add(self, item, index=None):
```

martinRenou (commented):
Looks like a nice addition! Would you be able to add some ui-tests for this? 

Also, I would rename `position` into `index`, what do you think?

gjmooney:
I added the UI tests, I also added the collapsed option to the LayerControl since that was easier then getting playwright to hover the mouse and take a screenshot

martinRenou:
> I added the UI tests

You probably forgot to commit the files for that? Only `leaflet.py` changed in the PR

gjmooney:
> > I added the UI tests
> 
> You probably forgot to commit the files for that? Only `leaflet.py` changed in the PR

lol woops

martinRenou:
update galata references

martinRenou (approved):
Thanks!

PR #1158, opened 2023-12-19
Ruff linting

gjmooney:
Move from Flake8 to Black for linting/formatting.


davidbrochart:
Ruff has a [formater](https://docs.astral.sh/ruff/formatter/) which is way faster than Black, I think it would be better.

examples/CanvasRenderer.ipynb
```diff
@@ -39,7 +39,7 @@
     "m = Map(center=(48.1, 17.1), zoom=8, prefer_canvas=True)\n",
     "\n",
     "gdf = geopandas.GeoDataFrame(\n",
-    "    geometry=[shapely.geometry.Point(l) for l in random_locs], crs=\"EPSG:4326\"\n",
+    "    geometry=[shapely.geometry.Point(l) for loc in random_locs], crs=\"EPSG:4326\"\n",
```
martinRenou, line 42:
```suggestion
    "    geometry=[shapely.geometry.Point(loc) for loc in random_locs], crs=\"EPSG:4326\"\n",
```

martinRenou (changes requested):

examples/CanvasRenderer.ipynb
gjmooney: Fixed and pushed

PR #1160, opened 2023-12-22
Start adding types

gjmooney:


js/src/controls/Control.ts
```diff
@@ -1,13 +1,20 @@
...
-
-import * as widgets from '@jupyter-widgets/base';
+import { WidgetModel, WidgetView } from '@jupyter-widgets/base';
 import L from '../leaflet';
-import * as utils from '../utils';
-export class LeafletControlModel extends widgets.WidgetModel {
-  defaults() {
+import { LeafletWidgetView } from '../utils';
+
+interface LeafletControlModelOptions {
```
martinRenou, line 8: We may want to export this as it will probably be useful in other places 
```suggestion
export interface LeafletControlModelOptions {
```

js/src/controls/Control.ts
```diff
@@ -40,14 +52,11 @@ export class LeafletControlView extends utils.LeafletWidgetView {
...
-        this.model,
-        'change:' + key,
-        function () {
-          L.setOptions(this.obj, this.get_options());
-        },
-        this
-      );
+      this.listenTo(this.model, 'change:' + key, function () {
+        L.setOptions(this.obj, this.get_options());
+      });
```
martinRenou, line 57: Using an arrow function so that the `this` context gets passed properly

```suggestion
      this.listenTo(this.model, 'change:' + key, () => {
        L.setOptions(this.obj, this.get_options());
      });
```

martinRenou (commented):
Looks good so far! 🚀  Thanks

js/src/Map.ts
```diff
@@ -100,119 +119,131 @@ export class LeafletMapModel extends widgets.DOMWidgetModel {
...
     super.initialize(options);
     // The dirty flag is used to prevent sub-pixel center changes
     // computed by leaflet to be applied to the model.
     this.dirty = false;
   }
 
   create_panes() {
-    const panes = this.model.get('panes');
+    // const panes = this.model.get('panes');
+    const panes = this.obj.getPanes();
```
martinRenou, line 204: I'm not sure about the diff here? We should probably keep the `model.get('panes')`. According to the git blame, this "panes" feature was introduced in [this PR](https://github.com/jupyter-widgets/ipyleaflet/pull/999) when fixing [this issue](https://github.com/jupyter-widgets/ipyleaflet/issues/993). You may test this feature running the `examples/MapPanes.ipynb` example.

js/src/Map.ts
```diff
@@ -100,119 +119,131 @@ export class LeafletMapModel extends widgets.DOMWidgetModel {
...
     for (const name in panes) {
       const pane = this.obj.createPane(name);
       const styles = panes[name];
       for (const key in styles) {
+        // TODO come back to this
+        //@ts-ignore
```
martinRenou, line 210: This may be removed once you resolve the comment about panes.

js/src/Map.ts
```diff
@@ -252,11 +283,11 @@ export class LeafletMapView extends utils.LeafletDOMWidgetView {
     });
   }
 
-  create_obj() {
+  async create_obj() {
```
martinRenou, line 286: More async functions 💯

js/src/Map.ts
```diff
@@ -321,133 +353,90 @@ export class LeafletMapView extends utils.LeafletDOMWidgetView {
...
-      'change:center',
-      function () {
-        if (!this.dirty) {
-          this.dirty = true;
-          this.obj.panTo(this.model.get('center'));
-          this.dirty = false;
-        }
-        this.model.update_bounds().then(() => {
-          this.touch();
+    // this.listenTo(this.model, 'msg:custom', this.handle_msg);
```
martinRenou, line 364: Let's remove this one

```suggestion
```

js/src/Map.ts
```diff
@@ -321,133 +353,90 @@ export class LeafletMapView extends utils.LeafletDOMWidgetView {
...
-      case 'foo':
-        break;
-    }
-  }
+  // handle_msg(content) {
+  //   switch (content.method) {
+  //     case 'foo':
+  //       break;
+  //   }
+  // }
```
martinRenou, line 427: WAT, let's remove this 😅 

It may have been some placeholder code that was never used, sorry for that.

```suggestion
```

js/src/Map.ts
```diff
@@ -321,133 +353,90 @@ export class LeafletMapView extends utils.LeafletDOMWidgetView {
...
 
-  processPhosphorMessage(msg) {
+  processPhosphorMessage(msg: Message) {
+    //TODO: fix this?
+    //@ts-ignore
```
martinRenou, line 431: This actually cannot compile (it uses APIs from the archived [Phosphor](https://github.com/phosphorjs/phosphor) library which is now [Lumino](https://github.com/jupyterlab/lumino)). It's used for JupyterLab 3 support.

So we should keep the ts-ignore and document it.

```suggestion
    // @ts-ignore This is for backward compatibility with JupyterLab 3
```

js/src/controls/Control.ts
```diff
@@ -20,34 +24,35 @@ export class LeafletControlModel extends widgets.WidgetModel {
...
     this.map_view = this.options.map_view;
   }
 
-  render() {
-    return Promise.resolve(this.create_obj()).then(() => {
-      this.leaflet_events();
-      this.model_events();
-    });
+  async render() {
+    await Promise.resolve(this.create_obj());
```
martinRenou, line 39: I wonder, can we do this? I guess not with an abstract method?

```suggestion
    await this.create_obj();
```

martinRenou (commented):
Super nice work!! Thanks!

js/src/leaflet.ts
```diff
@@ -51,25 +48,11 @@ L.patchGridLayer = function (layer: any) {
...
-  var obj = oldTileLayer(url, options);
-  L.patchGridLayer(obj);
-  return obj;
-};
-
-L.tileLayer.wms = function (url: any, options: any) {
-  var obj = oldTileLayer.wms(url, options);
-  L.patchGridLayer(obj);
-  return obj;
-};
```
martinRenou, line 73: We should keep these ugly hacks, you would need a ts-ignore around

js/src/leaflet-imageservice.ts
```diff
@@ -224,23 +229,27 @@ L.ImageService = L.Layer.extend({
...
+      const pixelorigin = this._topLeft.subtract(this._map.getPixelOrigin());
       L.DomUtil.setPosition(img, pixelorigin);
     }
   },
 
-  _update: function () {
+  // update: function () {
+  //   this.update();
+  // },
+
```
martinRenou, line 251:
```suggestion
```

js/src/layers/MagnifyingGlass.ts
```diff
@@ -24,53 +24,50 @@ export class LeafletMagnifyingGlassModel extends rasterlayer.LeafletRasterLayerM
...
-      var options = this.get_options();
-      options.layers = layers;
-      //@ts-ignore
-      this.obj = L.magnifyingGlass(options);
-    });
+
+    const layers = this.get_options().layers;
+    const layers_1 = await this.layer_views.update(layers);
+    const options = this.get_options();
+    options.layers = layers_1;
```
martinRenou, line 55:
```suggestion
    const layers_views = await this.layer_views.update(layers);
    const options = this.get_options();
    options.layers = layers_views
```

js/src/extension.ts
```diff
@@ -1,16 +1,4 @@
...
-  window.require.config({
-    map: {
-      '*': {
-        'jupyter-leaflet': 'nbextensions/jupyter-leaflet/index',
-      },
-    },
-  });
-}
-
-export const load_ipython_extension = function () {};
```
martinRenou, line 16: We need to keep this for jupyter notebook v6 support

js/src/amd-public-path.ts
```diff
@@ -1,10 +0,0 @@
-// In an AMD module, we set the public path using the magic requirejs 'module' dependency
-// See https://github.com/requirejs/requirejs/wiki/Differences-between-the-simplified-CommonJS-wrapper-and-standard-AMD-define#module
-// Since 'module' is a requirejs magic module, we must include 'module' in the webpack externals configuration.
-// @ts-nocheck
-
-import * as module from 'module';
-var url = new URL(module.uri, document.location);
-// Using lastIndexOf('/')+1 gives us the empty string if there is no '/', so pathname becomes '/'
-url.pathname = url.pathname.slice(0, url.pathname.lastIndexOf('/') + 1);
-__webpack_public_path__ = `${url.origin}${url.pathname}`; // eslint-disable-line no-undef
```
martinRenou, line 10: We also need to keep this for jupyter notebook v6 support

martinRenou (changes requested):
A couple more comments and I think we're good to go! Thanks!

martinRenou (approved):
Thanks a lot for this huge work!

PR #1162, opened 2024-01-08
Setup hatch and jupyter-releaser

martinRenou:


martinRenou:
Putting this on hold for now. Either we force the same JS and Python version or we need a strategy to keep them separated.

Ideally we would package jupyter-leaflet separately (maybe export it from this repo?)

PR #1164, opened 2024-01-08
Leaflet update

gjmooney:
Updates Leaflet version. Introduces a [breaking change](https://github.com/Leaflet/Leaflet/blob/main/CHANGELOG.md#180-2022-04-18) in Map API.


martinRenou (approved):
LGTM! Thanks

PR #1171, opened 2024-01-17
Add workaround for MeasureControl map panning bug

### LINKED ISSUE #1076: Measure distances and area tool bugging?
The measure distances and area tool appears to have a bug when used in Jupyter notebook, has anyone else experienced an issue?

When I click on the map to begin a measurement the screen shifts on each click (a uniform amount) and the start of the intended measure point shifts? This video demonstrates the issue:

https://user-images.githubusercontent.com/93473831/203305954-f5b983f5-5e24-4f00-9526-261f1b4b33e5.mp4

This error occurs on the basemap, without having added additional layers, ruling out a projection issue.
I am trying to use ipyleaflet through leafmap, and so I also created a new environment containing only Leafmap, to confirm it was not a package conflict, and the error still occurs.

The use of the measure tool is integral to my use of ipyleaflet/leafmap, I thank you for any time, support and guidance you can offer.

gjmooney:
Add a workaround to fix #1076

### DIFF
--- a/python/jupyter_leaflet/src/controls/MeasureControl.ts
+++ b/python/jupyter_leaflet/src/controls/MeasureControl.ts
@@ -5,6 +5,22 @@
 import L from '../leaflet';
 import * as control from './Control';
 
+// Workaround for map panning bug
+// from https://github.com/ljagis/leaflet-measure/issues/171#issuecomment-1137483548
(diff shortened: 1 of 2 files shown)


martinRenou (approved):
Thanks!

Issue #1165, opened 2024-01-09
Geoman support as draw controls alternative?


maartenbreddels:
Hi,

We've been looking into the draw controls for ipyleaflet. One thing we were interested in (cc @mangecoeur ) was controlling the draw controls programmatically, using our own controls (not the builtin toolbar).
It seems this is not possible with the leaflet-draw plugin (bad design?), and the project seems dead:
 * https://github.com/Leaflet/Leaflet.draw/issues/1065 (no reply) 
 * https://github.com/Leaflet/Leaflet.draw/issues/965 (suggests using geoman)
 * https://github.com/Leaflet/Leaflet.draw/issues/1041 (suggests using geoman)

Most people suggest using [Geoman](https://www.geoman.io/) / https://github.com/geoman-io/leaflet-geoman in the above issues, which seems like a good alternative.
I always hate breaking changes, so I suggest we *add* Geoman draw controls to ipyleaflet, and keep the current draw controls as they are right now.
Will a PR for this likely be merged for ipyleaflet if we do the work?
cc @iisakkirotko @mariobuikhuizen

Regards,

Maarten

maartenbreddels:
Awesome, yes, we should mention in the docs why both are there and direct people to the new one.

martinRenou:
> I always hate breaking changes, so I suggest we add Geoman draw controls to ipyleaflet, and keep the current draw controls as they are right now.

Would we be able to keep the same Python API while using geoman instead of leaflet-draw in the front-end?

It seems that having leaflet-draw there could break ipyleaflet entirely. leaflet-draw makes use of the `initHook` feature of LeafletJS, which allows them to patch the creation of e.g. markers. This seems to cause some issues in some cases (I have difficulties spotting the exact cases for now). So not having leaflet-draw in the page may be good.

Also I'm wondering if those plugins will not collide.

maartenbreddels:
So, you are asking, can we replace leaflet-draw with geoman?

martinRenou:
Yes

iisakkirotko:
Hey @martinRenou and @maartenbreddels! I took a look at this a while ago, so here are some of my thoughts on the discussion:

> Would we be able to keep the same Python API while using geoman instead of leaflet-draw in the front-end?

The one thing where I don't think there is _at least_ feature parity is polyline, which I don't think Geoman has. Otherwise I don't foresee any issues with this that can't be overcome.

[...]

PR #1178, opened 2024-02-29
ci: add draw controls test

maartenbreddels:
This is using solara to run the test with a live frontend.

These things are difficult to test with galata, since we test user interaction.

In light of a future PR (see https://github.com/jupyter-widgets/ipyleaflet/issues/1165 for details) I think it's good to test one of the more complex widgets.


maartenbreddels:
Failure is most likely not due to this PR.

martinRenou:
Thanks Maarten! 

I'm a bit skeptical about adding a new way to do visual regression testing to the repo. I'd prefer we either stick to galata or we move entirely to using solara testing.

> These things are difficult to test with galata

you more or less have the same API with playwright

maartenbreddels:
Hi Martin,

> you more or less have the same API with playwright

Yes, but I've noticed pure python people having no idea what to do with the `ui-tests`, and have no idea how it works (it took me a while). Building tests that do interaction (like click a button) are basically a no-go for more Python people.

That said, I totally understand not wanting two systems. I could probably rewrite the notebook tests to use the solara test framework, but that would mean we don't have such a nice HTML report, since that is not available for Python yet: https://github.com/microsoft/playwright-python/issues/1296

Let me know what you think? Do you think its worth to give up the HTML report to have the option to do more complex UI tests from Python?

cheers,

Maarten

PR #1180, opened 2024-03-01
fix: Support for Python 3.8 was removed

iisakkirotko:
I believe this was done accidentally as a part of the recent larger refactor, since testing is still being done on Python 3.8, while it isn't officially supported.


martinRenou (approved):
Thanks! CI failure is unrelated, let's fix it separately

PR #1181, opened 2024-03-01
feat: introduce leaflet-geoman as an alternative to leaflet-draw

iisakkirotko:
This replaces the old leaflet-draw implementation with a leaflet-geoman based one, in accordance with https://github.com/jupyter-widgets/ipyleaflet/issues/1165. The change should be backwards compatible in that even though leaflet-draw is no longer used on the front end, the functionality is available identically from the python side.


iisakkirotko:
The remaining visual test failure seems unrelated, see the reported diffs:
dc964591a90a9bc7445db0487ddaea75ac69c5aa
e7db508791c7ed51a8cc959d715b88037391c0b8

maartenbreddels:
@giswqs do you use DrawControls a lot? I wonder if you are willing to test this PR. We have to break things.

giswqs:
Yes, I can give it a try. Do I Just pip install from the branch? Anything else I need to do to make sure the JS part is included?

giswqs:
It seems the position of the draw control is not working properly. In the example below, the DrawControl is supposed to be placed beneath the LayersControl, but instead it is being added above the ZoomControl. Same thing happens when adding the draw control to other position, e.g., `topright`. In other words, the draw control will always be placed at the top or bottom corner, ignoring other controls that have already been placed in the corner. 

```python
from ipyleaflet import Map, LayersControl, FullScreenControl, DrawControl, GeomanDrawControl

m = Map(center=[40, -100], zoom=4)
m.layout.height='600px'
m.scroll_wheel_zoom=True
m.add(FullScreenControl())
m.add(LayersControl(position='topleft'))
draw_control = DrawControl(position='topleft')
m.add(draw_control)
m
```

**New DrawControl**
image

**GeomanDrawControl**
image

**Old DrawControl**
image

giswqs:
When a drawn polygon has no more than four vertices, it returns a `polygon` type. When it has more than four vertices, it returns a `Polygon` type. The old draw control returns a `Polygon` type no matter how many vertices the polygon has. 

**New DrawControl**
image

image

**Old DrawControl**
image
image

giswqs:
Can't enter text when the text layer is initially created. Need to click on the edit layer button in order to add text. 

The JavaScript version can add enter text when the layer is added. No need to click on the edit layer button. 

https://www.geoman.io/docs/text-layer

```python
from ipyleaflet import Map, GeomanDrawControl

m = Map(center=(50, 354), zoom=5)

draw_control = GeomanDrawControl()
draw_control.text = {
    "textOptions": {
        "text":  "Geoman is fantastic! 🚀"
    }
}

m.add(draw_control)
m.layout.height = '600px'
m.scroll_wheel_zoom = True
m
```
Peek 2024-03-26 09-26

giswqs:
How to add the marker tool to the draw control? The following code does not work. 

```python
from ipyleaflet import Map, GeomanDrawControl

m = Map(center=(50, 354), zoom=5)
draw_control = GeomanDrawControl()
draw_control.marker = {
    "markerStyle": {
        "draggable": True
    }}
draw_control.polyline =  {
    "pathOptions": {
        "color": "#6bc2e5",
        "weight": 8,
        "opacity": 1.0
    }
}
m.add(draw_control)
m.layout.height = '600px'
m.scroll_wheel_zoom = True
m
```
image

iisakkirotko:
Thanks a lot for your feedback @giswqs! I'll take a look at the issues later this week.

iisakkirotko:
Hey @giswqs!

b8863cc4d8a1d056ae8c8729f37ac71ac5dba1ab should address all of your comments, with the exception of the issue with the differing capitalisation of "polygon". The `type` you mention there is consistent, but admittedly confusingly named - one instance is under "`feature.geometry.type`", while the other is under "`feature.properties.type`". The second one is used, among other things, to differentiate between different types of `Marker` layers (`TextMarker`, `CircleMarker`, etc). Maybe renaming this one is in order to avoid confusion. I'd love to hear your thoughts on what you think a suitable name would be!

maartenbreddels:
Hi all,

first of all, some credits, and thank you's:
Thanks to @mangecoeur from Planeto SA for sponsoring this feature. Planeto develops geospatial solutions for district energy planning ([planeto-energy.ch](http://planeto-energy.ch/)) and use ipyleaflet a lot. It's great to see companies funding open source projects which they use, kudos to them.

Iisakki: Thank you for the hard work you put into building this feature. A lot of energy went into it.

@giswqs Thank you for the thorough review, it has improved the quality and backwards compatibility a lot.

I talked to @martinRenou privately and expressed my concerns about breaking backward compatibility despite my best efforts to avoid that.
Given that leaflet.draw is still working, there is no strong reason to get rid (e.g. a security issue, or strong maintenance issue), we thought it would be better for now to:

1. Keep the old DrawControl, nothing changes or breaks for existing users.
2. Introduce the new GeomanDrawControl (this is done in this PR) and DrawControlCompat (rename what is now DrawControl in this PR).

In the future, when needed, if we do a major release, we can then:
 1. remove the old `DrawControl`
 2. rename `DrawControlCompat` to `DrawControl` and add `DrawControlCompat` as an alias for backward compatibility.


I hope everyone agrees that's a safe and frictionless path forward.

cheers,

Maarten

martinRenou:
Thanks a lot for that work! ipyleaflet is really missing contributors right now. I personally don't have time/energy to continue any maintenance on it, but I'm happy to make releases from time to time.

Is the PR ready? Happy to give it a final review/merge.

iisakkirotko:
Hi @martinRenou!

I'll make this PR reflect the approach @maartenbreddels outlined in https://github.com/jupyter-widgets/ipyleaflet/pull/1181#issuecomment-2046786537 today, and open a draft PR for the future breaking change. So I expect this to be fully ready for review later today!

iisakkirotko:
@martinRenou PR should be ready for final review! The installation errors seem to be due to an upgrade to `@types/leaflet@1.9.9`, which should be fixed in https://github.com/jupyter-widgets/ipyleaflet/pull/1186

martinRenou:
Thanks! I just rebased from the UI, you'll need to `pull` again

python/jupyter_leaflet/src/Map.ts
```diff
@@ -234,6 +235,14 @@ export class LeafletMapView extends LeafletDOMWidgetView {
     const view = await this.create_child_view<LeafletControlView>(child_model, {
       map_view: this,
     });
+    if (
+      view instanceof LeafletGeomanDrawControlView &&
```
martinRenou, line 239: I don't like much making a special case here. We may want to find a better approach.

Though let's not consider this a blocker for merging the PR. Can you at least add a comment e.g. "TODO: Remove this special case"

martinRenou (commented):
Thanks!

python/jupyter_leaflet/src/Map.ts
iisakkirotko: I totally agree with you. Unfortunately geoman constructs and adds it's toolbar in it's own way, which means if we want to not have the special case here, it looks like we would have to reconstruct the whole toolbar ourselves.

martinRenou: Thanks for adding a comment 👌🏽  

Other question, if you remove the control Python side, does it work as intended? We may need a special case in the removal logic too?

iisakkirotko: Removal should work correctly now, although (at least as far as I can see) we don't need a special case, since we can just do this in the `remove` method of the class. Let me know if you foresee any issues with doing it this way!

iisakkirotko: Only geoman should add stuff there, since `L.PM` is added by geoman itself (apparently before becoming `leaflet-geoman`, the package was known as `leaflet.pm`)

docs/controls/geoman_draw_control.rst
```diff
@@ -0,0 +1,73 @@
+Geoman Draw Control
+============
+
+``GeomanDrawControl`` allows one to draw various shapes on the map. 
+The drawing functionality opn the front-end is provided by `geoman <https://geoman.io/>`_.
```
maartenbreddels, line 5:
```suggestion
The drawing functionality on the front-end is provided by `geoman <https://geoman.io/>`_.
```

docs/controls/geoman_draw_control.rst
```diff
@@ -0,0 +1,73 @@
...
+
+The following shapes are supported:
+- marker
+- circleMarker
```
maartenbreddels, line 9:
```suggestion
- circlemarker
```
I think should should match the trait name

python/ipyleaflet_core/ipyleaflet/leaflet.py
```diff
@@ -2215,6 +2192,135 @@ def clear_markers(self):
...
+
+    Alternative drawing tools for drawing on the map provided by Leaflet-Geoman.
+    """
+
+    _view_name = Unicode("LeafletGeomanDrawControlView").tag(sync=True)
+    _model_name = Unicode("LeafletGeomanDrawControlModel").tag(sync=True)
+
+    # Current mode & shape
+    # valid values are: 'draw', 'edit', 'drag', 'remove', 'cut', 'rotate'
+    # for drawing, the tool can be added after ':' e.g. 'draw:Marker'
```
maartenbreddels, line 2236:
```suggestion
    # for drawing, the tool can be added after ':' e.g. 'draw:marker'
```
not sure if that should match the trait name

python/jupyter_leaflet/src/controls/GeomanDrawControl.ts
```diff
@@ -0,0 +1,528 @@
...
+  defaults() {
+    return {
+      ...super.defaults(),
+      _view_name: 'LeafletGeomanDrawControlView',
+      _model_name: 'LeafletGeomanDrawControlModel',
+      current_mode: null,
+      hide_controls: false,
+      data: [],
+      marker: {},
+      circleMarker: { markerStyle: {} },
```
maartenbreddels, line 20:
```suggestion
      circlemarker: { markerStyle: {} },
```
should match trait name

python/jupyter_leaflet/src/controls/GeomanDrawControl.ts
```diff
@@ -0,0 +1,528 @@
...
+      circle: {},
+      polyline: { pathOptions: {} },
```
maartenbreddels, line 22: default options do not match the traitlet defau

python/ipyleaflet_core/ipyleaflet/leaflet.py
```diff
@@ -2215,6 +2192,135 @@ def clear_markers(self):
...
+    current_mode = Any(allow_none=True, default_value=None).tag(sync=True)
+
+    # Hides toolbar
+    hide_controls = Bool(False).tag(sync=True)
+
+    circlemarker = Dict({ 'shapeOptions': {} }).tag(sync=True)
```
maartenbreddels, line 2242: I think this can go into the base class.

python/ipyleaflet_core/ipyleaflet/leaflet.py
```diff
@@ -2215,6 +2192,135 @@ def clear_markers(self):
...
+class DrawControlCompatibility(DrawControlBase):
+    """DrawControl class.
+
+    Python side compatibility layer for old DrawControls, using the new Geoman front-end but old Python API.
+    """
+    
+    _view_name = Unicode("LeafletGeomanDrawControlView").tag(sync=True)
+    _model_name = Unicode("LeafletGeomanDrawControlModel").tag(sync=True)
+
+    circlemarker = Dict().tag(sync=True)
```
maartenbreddels, line 2296: In the default value should match what it was originally, this can go into the base class (if not, ignore my other comment on the same property in the Geoman class)

maartenbreddels (commented):
Just a few small issues.
Btw, I love how the compatibility part is only on the Python side.

python/ipyleaflet_core/ipyleaflet/leaflet.py
iisakkirotko: Indeed this should match the original case. However I've put them separately for the different classes, since for the geoman implementation the default options should be `'pathOptions': {}`, while for the other two they should be `'shapeOptions': {}` (although for the geoman implementations either will work).

maartenbreddels (approved):
@martinRenou do you think we can merge this?

martinRenou (approved):
Thanks!

PR #1186, opened 2024-04-11
fix: cast this.get_options to support @types/leaflet@1.9.9

iisakkirotko:
Fixes two issues encountered during typescript compilation with newest version of `@types/leaflet`, `1.9.9`.


martinRenou (approved):
Thanks!

martinRenou:
I'm sure UI tests failure are unrelated. Let's continue with this.

PR #1189, opened 2024-04-23
Refactor: remove ipyleaflet-core and have only ipyleaflet and jupyter-leaflet packages

martinRenou:
This will ease the releasing so much. ipyleaflet-core itself was not really useful on its own anyway.


python/ipyleaflet/pyproject.toml
```diff
@@ -29,12 +28,30 @@ classifiers = [
...
 requires-python = ">=3.8"
 dependencies = [
-    "ipyleaflet_core",
-    "jupyter_leaflet",
+    "ipywidgets>=7.6.0,<9",
+    "traittypes>=0.2.1,<3",
+    "xyzservices>=2021.8.1",
+    "branca>=0.5.0",
+    # Is there no automatic way to compute this with hatch??
+    "jupyter_leaflet>=0.19,<0.20"
```
martinRenou, line 36: Would love to resolve that at some point. But let's not make it a blocker for the PR.

PR #1197, opened 2024-05-14
fix(docs): GeomanDrawControl was not included in docs navigation

iisakkirotko:
The documentation for `GeomanDrawControl` was not linked to from the navigation before.


martinRenou (approved):
Thanks!

Issue #1198, opened 2024-05-23
Cannot edit display_options in Velocity


usereight8:
I'm trying to change the velocity units in ipyleaflet's Velocity but it seems that any change in the display_options dictionary is ineffective.

`from ipyleaflet import Map, TileLayer, basemaps
from ipyleaflet.velocity import Velocity
import xarray as xr
import os

if not os.path.exists('wind-global.nc'):
    url = 'https://github.com/benbovy/xvelmap/raw/master/notebooks/wind-global.nc'
    import requests
    r = requests.get(url)
    wind_data = r.content
    with open('wind-global.nc', 'wb') as f:
        f.write(wind_data)

center = [0, 0]
zoom = 1
m = Map(center=center, zoom=zoom, interpolation='nearest', basemap=basemaps.CartoDB.DarkMatter)

ds = xr.open_dataset('wind-global.nc')
display_options =         {
            "velocityType": "Global Wind",
            "position": "topleft",
            "emptyString": "No velocity data33",
            "angleConvention": "bearingCW",
            "displayPosition": "topleft",
            "displayEmptyString": "No velocity data33",
            "speedUnit": "k/h",
        }
wind = Velocity(data=ds,
                zonal_speed='u_wind',
                meridional_speed='v_wind',
                latitude_dimension='lat',
                longitude_dimension='lon',
                velocity_scale=0.01,
                max_velocity=20,
                display_options=display_options)
m.add(wind)

m`

I've tried editing the dictionary without any noticable changes. I've also tried some editing in the L.Control.Velocity.js file.

My main goal is to be able to display a different string rather than "m/s", as I can do the math conversion for the numbers before feeding them to the velocity module.

PR #1199, opened 2024-05-24
Fix velocity display_options bug

giswqs:
This PR fixes the bug reported in https://github.com/jupyter-widgets/ipyleaflet/issues/1198


martinRenou (approved):
Thanks! 👀

Issue #1195, opened 2024-05-02
ipyleaflet 0.19.0 can't be installed with nbextension


ancestor-mithril:
Version 0.18.0 works
```
pip install ipyleaflet==0.18.0
jupyter nbextension enable --py --sys-prefix ipyleaflet
```
The last release doesn't work anymore:
```
pip install ipyleaflet==0.19.0
jupyter nbextension enable --py --sys-prefix ipyleaflet
```
***
The error is :
```
Traceback (most recent call last):
  File "/opt/conda/bin/jupyter-nbextension", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/jupyter_core/application.py", line 283, in launch_instance
    super().launch_instance(argv=argv, **kwargs)
  File "/opt/conda/lib/python3.11/site-packages/traitlets/config/application.py", line 1075, in launch_instance
    app.start()
  File "/opt/conda/lib/python3.11/site-packages/notebook/nbextensions.py", line 972, in start
    super().start()
  File "/opt/conda/lib/python3.11/site-packages/jupyter_core/application.py", line 270, in start
    self.subapp.start()
  File "/opt/conda/lib/python3.11/site-packages/notebook/nbextensions.py", line 882, in start
    self.toggle_nbextension_python(self.extra_args[0])
  File "/opt/conda/lib/python3.11/site-packages/notebook/nbextensions.py", line 855, in toggle_nbextension_python
    return toggle(module,
           ^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/notebook/nbextensions.py", line 470, in enable_nbextension_python
    return _set_nbextension_state_python(True, module, user, sys_prefix,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/notebook/nbextensions.py", line 368, in _set_nbextension_state_python
    m, nbexts = _get_nbextension_metadata(module)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/notebook/nbextensions.py", line 1109, in _get_nbextension_metadata
    raise KeyError(
[...]

PR #1200, opened 2024-05-29
remove broken, deprecated nbextension enable from installation docs

minrk:
install docs still suggest

```
jupyter nbextension enable --py --sys-prefix ipyleaflet
```

which doesn't work anymore because nbextension methods moved to `jupyter_leaflet` without deprecation in #1189. It's unclear from #1189 whether breaking the nbextension name was intentional.

xref #1195

### DIFF
--- a/docs/installation/index.rst
+++ b/docs/installation/index.rst
@@ -7,7 +7,6 @@ Using pip
 .. code:: bash
 
     pip install ipyleaflet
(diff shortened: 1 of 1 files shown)


martinRenou (approved):
Thanks!

This is something we missed when splitting the ipyleaflet package in two

Issue #744, opened 2020-11-04
Add style callbacks for VectorTileLayer


deeplook:
Styling for vector tiles can be pretty complex and the current implementation of `VectorTileLayer` only allows for one static style per "tile layer" e.g. "roads". This is a severe limitation as one cannot exploit the rich feature properties stored in vector tiles, e.g. to render different types of roads individually. The Leaflet VectorGrid plugin does support function callbacks for more dynamic styling as described here: https://leaflet.github.io/Leaflet.VectorGrid/vectorgrid-api-docs.html#styling-vectorgrids. So I think it should be possible to wrap this functionality and expose it in ipyleaflet, too, in a way similar to the `style_callback` parameter for `GeoJSON` layer.

martinRenou:
I don't this is technically doable without transpiling the Python code into JS.

In GeoJSON we can modify the data with a `style_callback` because they are available in the Python process. In the VectorTileLayer we don't have access to the tiles data in the Python process, something like a `style_callback` would need to be performed in JS.

Also, we provide access to styling the VectorTileLayer already IIRC.

deeplook:
> In the VectorTileLayer we don't have access to the tiles data in the Python process

But the VectorGrid plugin has access to it or it couldn't apply the styling functions mentions on the page above. Maybe it's possible to have deeper access to this plugin?

> Also, we provide access to styling the VectorTileLayer already IIRC.

To some degree at least. Maybe @omanges can help.

martinRenou:
> But the VectorGrid plugin has access to it or it couldn't apply the styling functions mentions on the page above

Only JavaScript side.

A `style_callback` feature you are requesting would need to be provided in Python I guess. Unless we ask people to write JavaScript, but I personally don't wish anyone to have to write JavaScript.

deeplook:
> but I personally don't wish anyone to have to write JavaScript.

I surely share that feeling. ;) How about a JupyterLab extension, then? ;)

martinRenou:
> How about a JupyterLab extension, then? ;)

What do you mean? ipyleaflet already contains a JupyterLab extension

deeplook:
Sure, but maybe the extension could cover the more intricate JS code...

martinRenou:
What kind of API do you have in mind?

lopezvoliver:
> > But the VectorGrid plugin has access to it or it couldn't apply the styling functions mentions on the page above
> 
> Only JavaScript side.
> 
[...]

PR #1206, opened 2024-06-12
Improve vector tile layer

lopezvoliver:
This PR aims to add the ability to use dynamic styling to `VectorTileLayer`. That is, styling may be defined as a function of `properties` and `zoom`. This is possible with [Leaflet.VectorGrid](https://leaflet.github.io/Leaflet.VectorGrid/vectorgrid-api-docs.html), but is not yet implemented in ipyleaflet, because there is currently no way to define python callbacks that can be used by the leaflet map. 

One compromise is to allow the user to define the javascript code for the dynamic style as a string.

Following the [example from the documentation](https://ipyleaflet.readthedocs.io/en/latest/layers/vector_tile.html), here's a simple dynamic style that the user could define for use with `ipyleaflet.VectorTileLayer`: 

```python
jstyle='''{
    water:{opacity:0},
    places:{opacity:0},
    transit:{opacity:0},
    pois:{opacity:0},
    boundaries:{opacity:0},
    roads:{opacity:0},
    earth:{opacity:0},
    landuse: function(properties, zoom){
        var color = "gray";
        var fill = true;

        if(properties.kind=="forest"){
            color="green";
        }

        if(zoom<10){
            color="purple";
        }

        return {
            color: color,
            fill: fill
        }
        
    }
}
'''
```

it should only show `landuse` in purple for zoom levels less than 10, and otherwise show features as gray, except if they are "forest". 

Here's an animation of the behavior achieved by this PR:

ipyleaflet_new

Here's another example using my own vector tile layer:

ipyleaflet_new_pivots

Related issue:  #744

### DIFF
--- a/python/ipyleaflet/ipyleaflet/leaflet.py
+++ b/python/ipyleaflet/ipyleaflet/leaflet.py
@@ -1096,7 +1096,7 @@ class VectorTileLayer(Layer):
         Url to the vector tile service.
     attribution: string, default ""
         Vector tile service attribution.
-    vector_tile_layer_styles: dict or str, default {}
+    vector_tile_layer_styles: dict or string, default {}. If string, it will be parsed as a javascript object (useful for defining styles that depend on properties and/or zoom).
         CSS Styles to apply to the vector data.
     min_zoom: int, default 0
         The minimum zoom level down to which this layer will be displayed (inclusive).
--- a/python/ipyleaflet/ipyleaflet/leaflet.py
+++ b/python/ipyleaflet/ipyleaflet/leaflet.py
@@ -1114,6 +1114,8 @@ class VectorTileLayer(Layer):
     url = Unicode().tag(sync=True, o=True)
     attribution = Unicode().tag(sync=True, o=True)
 
+    vector_tile_layer_styles = Union([Dict(), Unicode()]).tag(sync=True, o=True)
+
     min_zoom = Int(0).tag(sync=True, o=True)
     max_zoom = Int(18).tag(sync=True, o=True)
     min_native_zoom = Int(default_value=None, allow_none=True).tag(sync=True, o=True)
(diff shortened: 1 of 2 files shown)


martinRenou:
Thanks for your PR!

Concerning the CI build failure, you should be able to fix it by running `yarn run lint` from the `python/jupyter_leaflet` directory

lopezvoliver:
@martinRenou thanks, that worked for the "run" tests. What about the UI / Visual regression test? is there something I can do?

python/jupyter_leaflet/src/layers/VectorTileLayer.ts
```diff
@@ -13,18 +13,39 @@ export class LeafletVectorTileLayerModel extends LeafletLayerModel {
...
-    const options = {
+  async create_obj() {
+    let options = {
       ...this.get_options(),
-      rendererFactory: L.canvas.tile,
     };
+    options['rendererFactory'] = L.canvas.tile;
+
+    let x: any = this.model.get('vectorTileLayerStyles');
+    if (typeof x !== 'object') {
```
martinRenou, line 34: Should it be?

```suggestion
    if (typeof x === 'string') {
```

python/ipyleaflet/ipyleaflet/leaflet.py
```diff
@@ -1096,8 +1096,16 @@ class VectorTileLayer(Layer):
         Url to the vector tile service.
     attribution: string, default ""
         Vector tile service attribution.
-    vector_tile_layer_styles: dict, default {}
+    vector_tile_layer_styles: dict or str, default {}
```
martinRenou, line 1099: Maybe adding a tiny bit of docstring for the case where it's a string?

martinRenou (commented):
Thanks! I believe the CI failures are unrelated

martinRenou (approved):
Thanks!

PR #1208, opened 2024-06-13
Add basemap model_id to Map._layer_ids attribute

giswqs:
When initializing the Map with a basemap, the basemap `model_id` is not added to the `Map._layer_ids` attribute, resulting in an error when trying to removing the basemap from the map. This PR fixes the bug by adding the basemap layer id to the `Map._layer_ids` attribute

```python
import ipyleaflet
m = ipyleaflet.Map()
m
m.remove(m.layers[0])
```

image


martinRenou (approved):
Thanks!

PR #1210, opened 2024-06-25
Vector tile layer opacity and visible

lopezvoliver:
This PR adds the missing `visible` and `opacity` traits/attributes in vector tile layers, so now it's possible to control the opacity and visibility of these layers using custom widgets.

Here's how it looks like using a custom widget with a checkbox (linked to `visible`) and a slider (linked to `opacity`):

ipyleaflet_opacity_visible

This solves the downstream leafmap [issue](https://github.com/opengeos/leafmap/issues/747) associated with its custom layer control toolbar:

leafmap_opacity_visible


martinRenou:
I believe the failures are unrelated..

Issue #1095, opened 2023-01-27
VectorTileLayer does not render when zoom level greater than 14


giswqs:
The ipyleaflet `VectorTileLayer` does not render when zoom level is greater than 14. I found the `maxNativeZoom` [issue](https://github.com/Leaflet/Leaflet.VectorGrid/issues/74) in the upstream [Leaflet.VectorGrid](https://github.com/Leaflet/Leaflet.VectorGrid/issues/74). Unfortunately, the repo is no longer being actively maintained and has no update since Sept 2021. It is unlikely that the issue will be resolved any time soon. 

There is an alternative called [`Leaflet.VectorTileLayer`](https://gitlab.com/jkuebart/Leaflet.VectorTileLayer), which is being actively developed by @jkuebart. It appears to support `min/maxNativeZoom`.

It would be great if ipyleaflet can switch to use `Leaflet.VectorTileLayer`.

The beauty of vector tile is that you can zoom way in to see the details of the shapes. Without the `maxNativeZoom` and `maxZoom`, the vector tile rendering is not very useful. 

```python
import ipyleaflet
m = ipyleaflet.Map(center=(40, -100), zoom=4, scroll_wheel_zoom=True)
url = 'https://planetarycomputer.microsoft.com/api/data/v1/vector/collections/ms-buildings/tilesets/global-footprints/tiles/{z}/{x}/{y}'
layer = ipyleaflet.VectorTileLayer(url=url, attribution='Microsoft', max_native_zoom=20, max_zoom=20)
m.add(layer)
m
```

![Peek 2023-01-27 09-13](https://user-images.githubusercontent.com/5016453/215107809-3783a949-7d21-4090-bc5c-634512b1c0ad.gif)

Relevant issues:

- https://github.com/microsoft/PlanetaryComputer/discussions/175
- https://github.com/python-visualization/folium/issues/1715

lopezvoliver:
@giswqs There are a couple of separate issues here. 

First, the current implementation of VectorTileLayer in ipyleaflet does not have the `max_native_zoom` parameter implemented. So, if you zoom in more than the actual max native zoom of a dataset, the layer doesn't appear. In #1206, I added this parameter, so this will work soon. 

[...]

PR #1213, opened 2024-06-28
Interactive vector tile layers

lopezvoliver:
## Renderer factory

Vector tile layers can be [rendered](https://leaflet.github.io/Leaflet.VectorGrid/vectorgrid-api-docs.html#svg-vs-canvas) using either `L.canvas.tile` or `L.svg.tile`. Currently, only the canvas option is implemented in ipyleaflet and it's not doing a great job at higher levels of zoom, see for example #1095. 

Here a new `renderer` option was added to `VectorTileLayer`, which can be either `svg` or `canvas`, with the new default being `svg`. 

Here's an example using the `ms-buildings` from Microsoft Planetary Computer, with the new default (svg):

```python
import ipyleaflet
m = ipyleaflet.Map(center=(41.91867,-88.10602), zoom=15)
url = 'https://planetarycomputer.microsoft.com/api/data/v1/vector/collections/ms-buildings/tilesets/global-footprints/tiles/{z}/{x}/{y}'
layer = ipyleaflet.VectorTileLayer(
    url=url, 
    attribution='Microsoft', 
    max_native_zoom=13,
    max_zoom=20,
    vector_tile_layer_styles={"bingmlbuildings":{
        "fill":True,
        "weight":0.5
    }},    
)
m.add(layer)
m
```

And here's how it renders using `renderer='canvas'`:

## Interactivity

A new `interactive` option was added with `False` as default, which enables the user to add listeners to the layer, which include information about the feature. Note that the default `renderer='svg'` option should be used for interactivity. For example:

```python
m = ipyleaflet.Map()
layer = ipyleaflet.VectorTileLayer(
    url=url_data,  # My vector tile layer 
    interactive=True,  # New interactive option 
    max_native_zoom=13,
    max_zoom=20,
    renderer='svg', # New renderer option. Defaults to 'svg'
    layer_styles=jStyles,   # javascript function given as a string
    feature_id = 'label', # New feature_id option. Here, 'label' is the name of the (numeric) 
    # attribute in my layer that uniquely identifies each feature (see below for more information)
)
m.add(layer)

def handle_click(**kwargs):    
    if ("properties" in kwargs):
        properties = kwargs["properties"]
        options = kwargs["options"]
        print(properties)
        print(options)

layer.on_click(handle_click)  

m
```

### `feature_id`

This is an optional attribute that is used to construct a simple javascript function to uniquely identify a feature. This is required if you will be updating feature styles through the new `set_feature_style` and `reset_feature_style` methods. The javascript function is of the form:

```js
function (feat: any) {return feat.properties[idVar];}
```

where `feat` is the feature, and `idVar` is the name of the (numeric) attribute in the layer to identify a feature. Note that features with the same id will be treated as one when changing style (see the original `getFeatureId` [documentation here](https://leaflet.github.io/Leaflet.VectorGrid/vectorgrid-api-docs.html#updating-styles)).

### Updating styles

Two new methods for `VectorTileLayer` were added: `set_feature_style` and `reset_feature_style`. The first one is used to update the style for an individual feature, which is useful for highlighting a feature (e.g., on click or mouseover), while the second one is useful for resetting the style to the default (e.g. to clear the highlight). 

### Example

Here's a motivating example that demonstrates all of the new features. 

```python
m = ipyleaflet.Map()
layer = ipyleaflet.VectorTileLayer(
    url=url_data,  # My vector tile layer 
    interactive=True,  # New interactive option 
    max_native_zoom=13,
    max_zoom=20,
    renderer='svg', # New renderer option. Defaults to 'svg'
    layer_styles=jStyles,   # javascript function given as a string
    feature_id = 'label', # New feature_id option. Here, 'label' is the name of the (numeric) 
    # attribute in my layer that uniquely identifies each feature.
)
m.add(layer)

info_title = "<h4>Field info</h4>"
info_default_value = "Hover over a field" 

info_widget = widgets.HTML(value=info_title + info_default_value)

highlight_style = {
                "weight": 5,
                "color": '#666',
                "dashArray": '',
                "fillOpacity": 0.7,
                "fill": True 
                }

def highlight_feature(**kwargs):    
    if ("properties" in kwargs):
        properties = kwargs["properties"]
        options = kwargs["options"]
        feature_id = properties["label"]
        fill_color = options["fillColor"]
        highlight_style["fillColor"] = fill_color
        layer.set_feature_style(
            id=feature_id,
            layer_style=highlight_style,
        )
        info_html = info_title
        for k,v in properties.items():
            info_html += "<b>"+ k + "</b>" + ": " + str(v) + "<br />"
        info_widget.value = info_html

def clear_highlight(**kwargs):
    if layer.feature_style: 
        layer.reset_feature_style(layer.feature_style["id"])
        info_widget.value = info_title + info_default_value
        
layer.on_mouseover(highlight_feature)
layer.on_mouseout(clear_highlight)

widget_control = ipyleaflet.WidgetControl(widget=info_widget, position='topright')
m.add(widget_control)

m
```

### DIFF
--- a/python/ipyleaflet/ipyleaflet/leaflet.py
+++ b/python/ipyleaflet/ipyleaflet/leaflet.py
@@ -1114,7 +1114,7 @@ class VectorTileLayer(Layer):
         Engine for rendering VectorTileLayers; either 'canvas' or 'svg'. Use 'svg' for interactive layers.
     interactive: boolean, default False
         Whether the layer is interactive or not.
-    get_feature_id: string, default None
+    feature_id: string, default None
         Optional attribute name of a unique feature identifier. 
     """
 
@@ -1133,7 +1133,7 @@ class VectorTileLayer(Layer):
     min_native_zoom = Int(default_value=None, allow_none=True).tag(sync=True, o=True)
     max_native_zoom = Int(default_value=None, allow_none=True).tag(sync=True, o=True)
     renderer = Unicode('svg').tag(sync=True, o=True)
-    get_feature_id = Unicode(allow_none=True, default_value=None).tag(sync=True, o=True)
+    feature_id = Unicode(allow_none=True, default_value=None).tag(sync=True, o=True)
     feature_style = Dict().tag(sync=True)
 
     def redraw(self):
--- a/python/ipyleaflet/ipyleaflet/leaflet.py
+++ b/python/ipyleaflet/ipyleaflet/leaflet.py
@@ -1110,7 +1110,8 @@ class VectorTileLayer(Layer):
         Opacity of the layer between 0. (fully transparent) and 1. (fully opaque).
     visible: boolean, default True
         Whether the layer is visible or not.
-    renderer_factory: rendererFactory option ('svg' or 'canvas') for L.VectorGrid, default 'svg'
+    renderer_factory: string, default 'svg'
+        Engine for rendering VectorTileLayers; either 'canvas' or 'svg'. Use 'svg' for interactive layers.
     interactive: boolean, default False
         Whether the layer is interactive or not.
     get_feature_id: string, default None
@@ -1141,6 +1142,38 @@ class VectorTileLayer(Layer):
         need to refresh the layer.
         """
         self.send({"msg": "redraw"})
+
+    def set_feature_style(self, id:Int, layer_style:Dict):
+        """Re-symbolize one feature.
+
+        Given the unique ID for a vector features, re-symbolizes that feature across all tiles it appears in. 
+        Reverts the effects of a previous set_feature_style call. get_feature_id must be defined for 
+        set_feature_style to work. 
+
+        Attributes
+        ----------
+        id: int
+            The unique identifier for the feature to re-symbolize
+        layer_styles: dict
+            Style to apply to the feature
+        """
+        self.send({"msg": 
+                   {"setFeatureStyle":
+                    {"id":id, "layerStyle":layer_style}}
+                })
+
+    def reset_feature_style(self, id:Int):
+        """Reset feature style
+
+        Reverts the style to the layer's deafult.
+
+        Attributes
+        ----------
+        id: int
+            The unique identifier for the feature to re-symbolize
+        """
+        self.send({"msg": {"resetFeatureStyle":{"id":id}}})
+
 
 
 class PMTilesLayer(Layer):
--- a/python/ipyleaflet/ipyleaflet/leaflet.py
+++ b/python/ipyleaflet/ipyleaflet/leaflet.py
@@ -1096,7 +1096,7 @@ class VectorTileLayer(Layer):
         Url to the vector tile service.
     attribution: string, default ""
         Vector tile service attribution.
-    vector_tile_layer_styles: dict or string, default {}. If string, it will be parsed as a javascript object (useful for defining styles that depend on properties and/or zoom).
+    layer_styles: dict or string, default {}. If string, it will be parsed as a javascript object (useful for defining styles that depend on properties and/or zoom).
         CSS Styles to apply to the vector data.
     min_zoom: int, default 0
         The minimum zoom level down to which this layer will be displayed (inclusive).
@@ -1124,7 +1124,7 @@ class VectorTileLayer(Layer):
     url = Unicode().tag(sync=True, o=True)
     attribution = Unicode().tag(sync=True, o=True)
 
-    vector_tile_layer_styles = Union([Dict(), Unicode()]).tag(sync=True, o=True)
+    layer_styles = Union([Dict(), Unicode()]).tag(sync=True, o=True)
     opacity = Float(1.0, min=0.0, max=1.0).tag(sync=True,o=True)
     visible = Bool(True).tag(sync=True, o=True)
     interactive = Bool(False).tag(sync=True, o=True)
--- a/python/ipyleaflet/ipyleaflet/leaflet.py
+++ b/python/ipyleaflet/ipyleaflet/leaflet.py
@@ -1113,6 +1113,8 @@ class VectorTileLayer(Layer):
     renderer_factory: rendererFactory option ('svg' or 'canvas') for L.VectorGrid, default 'svg'
     interactive: boolean, default False
         Whether the layer is interactive or not.
+    get_feature_id: string, default None
+        Optional attribute name of a unique feature identifier. 
     """
 
     _view_name = Unicode("LeafletVectorTileLayerView").tag(sync=True)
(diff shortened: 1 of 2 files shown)


lopezvoliver:
@martinRenou could you please suggest someone for review?

martinRenou:
Thanks! I'll have a look

python/ipyleaflet/ipyleaflet/leaflet.py
```diff
@@ -1134,6 +1143,38 @@ def redraw(self):
...
+        set_feature_style to work. 
+
+        Attributes
+        ----------
+        id: int
+            The unique identifier for the feature to re-symbolize
+        layer_styles: dict
+            Style to apply to the feature
+        """
+        self.send({"msg":
```
martinRenou, line 1160: I'm wondering if this could be a traitlets property instead of a custom comm message?

Something like 
```python
feature_styles = Dict().tag(sync=True)
```

`feature_styles` being the {id: layerStyle} dictionary?

Custom comm messages are transient and will not be saved in your widget model, resulting in many issues.

lopezvoliver: How would that work in `VectorTileLayer.ts` so that the `this.obj.setFeatureStyle` is called when the python method is called?

martinRenou: You would need to set a new value for `feature_styles` Python side:

```python
def set_feature_style(self, id:Int, layer_style:Dict):
    self.feature_styles = {**feature_styles, id: layer_style}
```

Then you would listen for changes in the JavaScript side:

```python
this.model.on("change:feature_styles", () => {
  const feature_styles = this.model.get("feature_styles");
  // You may need to reset all previous features styles here
  // [...]
  for (const id of Object.keys(feature_styles)) {
    this.obj.setFeatureStyle(id, feature_styles[id]);
  }
})
```

martinRenou:
> Here's an example using the ms-buildings from Microsoft Planetary Computer, with the new default (svg):

@lopezvoliver I was curious to run this example, but it seems I'm getting 404s on this https://planetarycomputer.microsoft.com url when LeafletJS tries to download the tiles. Is there anything particular to do to allow this example to work?

lopezvoliver:
> > Here's an example using the ms-buildings from Microsoft Planetary Computer, with the new default (svg):
> 
> @lopezvoliver I was curious to run this example, but it seems I'm getting 404s on this https://planetarycomputer.microsoft.com url when LeafletJS tries to download the tiles. Is there anything particular to do to allow this example to work?

I just checked and it should work as is. Are you able to access the following tile directly?

https://planetarycomputer.microsoft.com/api/data/v1/vector/collections/ms-buildings/tilesets/global-footprints/tiles/13/2090/3043

This layer is only available until z=13, so the `maxNativeZoom` (added in #1206) is set to 13.

python/ipyleaflet/ipyleaflet/leaflet.py
lopezvoliver: Thanks, I'll give it a try and make a new commit. 

FYI I am also testing another commit with changes to how `rendererFactory`, `getFeatureId`, and `vectorTileLayerStyles` are exposed in python, because I encountered an issue as follows:

1. Create a vector tile layer with `renderer_factory='svg'`, which is a string. 
2. In javascript, the `rendererFactory` is set to a function `L.svg.tile`.
3. Change the opacity in python.
4. Move the map
5. Error in javascript -- because renderer_factory is in sync with `rendererFactory`, it's now set to the string `svg`, and it throws a javascript error that the `rendererFactory` is not a function.

The new change to avoid this issue is that in python, instead of `rendererFactory` I will have a separate option called `renderer`, which is used in javascript to create `rendererFactory`. A similar strategy can be used for `getFeatureId` and `vectorTileLayerStyles` because they are constructed in javascript as functions, they shouldn't be in 'sync' with python. 

I will aim to commit this change soon as well.

lopezvoliver:
Here's a description of the recent five commits:

1. As suggested, I used `feature_style` as a traitlets property so that we can add observe its change in javascript and call the `setFeatureStyle` method. The user can call the `set_feature_style` method with an id and a `layer_style` dictionary. The user can also call the `reset_feature_style` method with an id, which updates the same `feature_style` traitlets property but with a `reset` value set to True. At this time, only one id should be used at a time, as is the intended use case of the original `setFeatureStyle` and `resetFeatureStyle` methods in VectorGrid. 

2. I realized that in #1210 one small issue persisted: when initializing a VectorTileLayer with `visible=False`, this wasn't reflected initially. This commit fixes this issue by setting `opacity=0` after initializing the vector tile layer with `visible=False`. 

3, 4, and 5: Each of these commits is a decision to rename `vector_tile_layer_styles`, `renderer_factory`, and `get_feature_id` (the latter two are additions in this pull request) to a different name in python (`layer_styles`, `renderer`, and `feature_id`, respectively). The reason for this is that because all three are converted to functions in javascript, they shouldn't be synced to the python model. Instead, they are used to construct the javascript counterparts `VectorTileLayerStyles`, `rendererFactory`, and `getFeatureId`.  

I will update the original PR comment to reflect these changes where needed.

martinRenou:
> This layer is only available until z=13, so the maxNativeZoom (added in https://github.com/jupyter-widgets/ipyleaflet/pull/1206) is set to 13.

Ah right I was not using the main branch! Thanks for the clarification.

> Here's a description of the recent five commits:

Is it ready for another review? 

Happy to make a release of your recent PRs after this one is merged

lopezvoliver:
> Is it ready for another review?

Yes, it's ready for review.

> Happy to make a release of your recent PRs after this one is merged

That would be great, thanks!

python/ipyleaflet/ipyleaflet/leaflet.py
```diff
@@ -1118,13 +1124,25 @@ class VectorTileLayer(Layer):
...
     max_zoom = Int(18).tag(sync=True, o=True)
     min_native_zoom = Int(default_value=None, allow_none=True).tag(sync=True, o=True)
     max_native_zoom = Int(default_value=None, allow_none=True).tag(sync=True, o=True)
+    renderer = Unicode('svg').tag(sync=True, o=True)
+    feature_id = Unicode(allow_none=True, default_value=None).tag(sync=True, o=True)
+    feature_style = Dict().tag(sync=True)
+
+    def __init__(self, **kwargs):
+        super(VectorTileLayer, self).__init__(**kwargs)
+        # Backwards compatibility: allow vector_tile_layer_styles as input:
```
martinRenou, line 1142: This will be backward compatible only at the layer creation.

I see three options here:
- should we really rename this property? If not we can keep `vector_tile_layer_styles`?
If yes:
- we remain backward compatible and we'd need a Python setter too `@vector_tile_layer_styles.setter def set_vector_tile_layer_styles`
- we remove backward compatibility and go with 0.20.0

I'd prefer one of the two first options

lopezvoliver: @martinRenou 

I went ahead with option 2: `vector_tile_layer_styles` is backwards compatible, and it uses a setter that updates `layer_styles`. 

Regarding the ability to update styles after initialization -- this wasn't contemplated in #1206, so I also fixed this by adding a listener to `layer_styles`. 

Here's a demo of the new capabilities:

martinRenou (approved):
Thanks!

martinRenou:
The CI would deserve some love, let's do that separately

PR #1214, opened 2024-07-04
Add on_zoom

lopezvoliver:
This new `on_zoom` feature is useful for implementing zoom-dependent logic, e.g. adding or removing components to the map based on zoom level. 

For example:

```python
def handle_zoom(**kwargs):
    z = kwargs["zoom"]
    legend_on_map = legend_control.model_id in m._control_ids
    if (z>9):
        if not legend_on_map:
            m.add(legend_control)
    else:
        if legend_on_map:
            m.remove(legend_control)

m.on_zoom(handle_zoom)
```

where `legend_control` is a control that I want shown only on zoom 10 or above:

### DIFF
--- a/python/ipyleaflet/ipyleaflet/leaflet.py
+++ b/python/ipyleaflet/ipyleaflet/leaflet.py
@@ -3171,13 +3171,19 @@ class Map(DOMWidget, InteractMixin):
 
     # Event handling
     _interaction_callbacks = Instance(CallbackDispatcher, ())
+    _zoom_callbacks = Instance(CallbackDispatcher, ())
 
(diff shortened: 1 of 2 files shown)


martinRenou:
Thanks! I believe this was already possible using traitlets:
```python
m = Map()

m.observe(handle_zoom, 'zoom')
```

lopezvoliver:
> Thanks! I believe this was already possible using traitlets:
> 
> ```python
> m = Map()
> 
> m.observe(handle_zoom, 'zoom')
> ```

This works, I can access the 'new' zoom in my `handle_zoom` function. Thanks!

PR #1225, opened 2024-08-18
Add scrollWheelZoom.enable() and disable()

### LINKED ISSUE #1224: Updating `scroll_wheel_zoom` doesn't affect existing maps.
Updating `scroll_wheel_zoom` does not enable/disable the corresponding behavior on an existing map.

Example: `scroll_wheel_zoom` is set to True initially, then the user updates it to `False`, but the map is still scrollable:

ipyleaflet-bug-scroll

lopezvoliver:
This pull request fixes #1224 

Update: also fixes the same issue with other types of zoom:

- doubleClickZoom
- touchZoom
- boxZoom

### DIFF
--- a/python/jupyter_leaflet/src/Map.ts
+++ b/python/jupyter_leaflet/src/Map.ts
@@ -423,6 +423,14 @@ export class LeafletMapView extends LeafletDOMWidgetView {
         this.obj.toggleFullscreen();
       }
     });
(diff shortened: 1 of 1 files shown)


lopezvoliver:
@martinRenou, could you please have a look?

martinRenou (approved):
Thanks!

PR #1241, opened 2025-01-11
Preserve `{r}` in basemap urls

letmaik:
Reference: https://xyzservices.readthedocs.io/en/stable/api.html#xyzservices.TileProvider.build_url

Without this PR, the only way to get 2x base maps is to manually instantiate a `TileLayer`.


arjxn-py (approved):
Thanks @letmaik

PR #1247, opened 2025-04-23
Fix geoman controls option update + snapping option

mangecoeur:
- Currently, if options are changed on the GeomanDrawControl after creation, the toolbar doesn't update. This PR adds event listeners to the model attributes to remove & re-add the draw toolbar on options change so they are reflected in the UI.
- This PR also adds the snapping option to Layers, to control which layers are used by the snapping function of Geoman.
- The version of Geoman is bumped to the latest version.


python/jupyter_leaflet/src/controls/GeomanDrawControl.ts
```diff
@@ -57,26 +60,32 @@ export class LeafletGeomanDrawControlView extends LeafletControlView {
...
-          options = feature.properties.options;
+          options = feature.properties?.options;
         }
         switch (feature.properties.type) {
           case 'marker':
+            if (!options) {
+              options = model.get('marker')?.markerStyle;
+            }
             return new L.Marker(latlng, options);
           case 'circle':
```
arjxn-py, line 78:
```suggestion
          case 'circle':
          	if (!options) {
              options = model.get('circle')?.pathOptions;
            }
```

When i was testing it, I got options `undefined` everytime, so let's add a check like other cases?

python/jupyter_leaflet/src/controls/GeomanDrawControl.ts
```diff
@@ -525,4 +539,47 @@ export class LeafletGeomanDrawControlView extends LeafletControlView {
...
+    const excluded_keys = ['current_mode'];
+    for (let key in this.model.attributes) {
+      if (!(key.startsWith('_') || excluded_keys.includes(key))) {
+        this.listenTo(this.model, 'change:' + key, () => {
+          this.setControlOptions();
+
+          this.map_view.obj.pm.removeControls();
+          if (!this.model.get('hide_controls')) {
+            this.map_view.obj.pm.addControls(this.controlOptions);
+          }
```
arjxn-py, line 556: This part of code is also being repeated in `setPosition()` below, maybe you want to take it out in a separate method of in you have any better idea?

arjxn-py:
Thanks @mangecoeur, this is awesome!!!

martinRenou (approved):
Thanks!

PR #1250, opened 2025-05-12
feat: add on_click and hover_style to GeomanDrawControl

lopezvoliver:
This PR adds an `on_click` method to GeomanDrawControl so that the user can define callbacks when clicking on the layers added using the GeomanDrawControl

Update: also added `hover_style`


arjxn-py (commented):
Thanks @lopezvoliver, this works pretty nicely :)


https://github.com/user-attachments/assets/59513eef-5fee-4824-82a9-24ebee199988




There are just a couple of minor suggestions: 

- You'd need to run `jlpm run lint` in `python/jupyter_leaflet` directory in order to make the CI happy
- Would you also want to update example on https://github.com/jupyter-widgets/ipyleaflet/blob/master/docs/controls/geoman_draw_control.rst a bit to showcase new methods added?

martinRenou:
These changes were included in https://github.com/jupyter-widgets/ipyleaflet/pull/1261 thanks a lot!!

PR #1251, opened 2025-05-13
fix: geojson layers pmIgnore

lopezvoliver:
Similar to #1249 , this PR updates jupyter_leaflet so that individual layers created by the LeafletGeoJSONModel get the proper `pmIgnore` option -- and disable the `pm` if it's already attached. This fixes the issue that Geoman is able to edit GeoJson layers, even though they inherit the `pm_ignore` from the base Layer class.

### DIFF
--- a/python/jupyter_leaflet/src/layers/GeoJSON.ts
+++ b/python/jupyter_leaflet/src/layers/GeoJSON.ts
@@ -63,6 +63,14 @@ export class LeafletGeoJSONView extends LeafletFeatureGroupView {
             coordinates: [e.latlng.lat, e.latlng.lng],
           });
         };
+        const pmIgnore = this.model.get('pm_ignore');
+        if (pmIgnore !== undefined) {
+          (layer as any).pmIgnore = pmIgnore;
+            if (pmIgnore && layer.pm) {
+              layer.pm.disable();
+              delete (layer as any).pm;
+            }
+        }
         layer.on({
           mouseover: mouseevent,
           click: mouseevent,
--- a/python/jupyter_leaflet/src/layers/GeoJSON.ts
+++ b/python/jupyter_leaflet/src/layers/GeoJSON.ts
@@ -66,10 +66,10 @@ export class LeafletGeoJSONView extends LeafletFeatureGroupView {
         const pmIgnore = this.model.get('pm_ignore');
         if (pmIgnore !== undefined) {
           (layer as any).pmIgnore = pmIgnore;
-            if (pmIgnore && layer.pm) {
-              layer.pm.disable();
-              delete (layer as any).pm;
-            }
+          if (pmIgnore && layer.pm) {
+            layer.pm.disable();
+            delete (layer as any).pm;
+          }
         }
         layer.on({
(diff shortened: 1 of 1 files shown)


mangecoeur:
There's probably overlap with this PR https://github.com/jupyter-widgets/ipyleaflet/pull/1247

lopezvoliver:
> There's probably overlap with this PR #1247

Hey @mangecoeur, I tested your fork but GeomanDrawControl was still able to modify GeoJSON layers:

bug_geojson_layer

I think the important part (in this PR) that solves this issue [is this](https://github.com/lopezvoliver/ipyleaflet/blob/geojson-pmignore/python/jupyter_leaflet/src/layers/GeoJSON.ts#L66-L69):

```typescript
            if (pmIgnore && layer.pm) {
              layer.pm.disable();
              delete (layer as any).pm;
            }
```

lopezvoliver:
> There's probably overlap with this PR #1247

Hey @mangecoeur, I tested an environment with PR #1247 and Geoman was still able to modify GeoJSON layers:

bug_geojson_layer

I think the important part of this PR that solves this issue [is this](https://github.com/lopezvoliver/ipyleaflet/blob/geojson-pmignore/python/jupyter_leaflet/src/layers/GeoJSON.ts#L66-L69):

```typescript
            if (pmIgnore && layer.pm) {
              layer.pm.disable();
              delete (layer as any).pm;
            }
```

mangecoeur:
Thanks, actually for my current use i mashed together both my current PR and your fixes into a kind of frakenpackage because I couldn't wait for a new ipyleaflet release. Would really like to land all of these fixes in the main repo instead.

lopezvoliver:
Same 😅 -- I have a custom mix and match of recent fixes. Hopefully the maintainers have some time to review these soon 👀

arjxn-py (commented):
Thanks @lopezvoliver 

You'd just need to run `jlpm run lint` in python/jupyter_leaflet directory in order to make the CI happy

And we'll also trigger ui-tests snapshots update to hopefully make ui-tests pass

python/jupyter_leaflet/src/layers/GeoJSON.ts
```diff
@@ -60,6 +60,14 @@ export class LeafletGeoJSONView extends LeafletFeatureGroupView {
             coordinates: [e.latlng.lat, e.latlng.lng],
           });
         };
+        const pmIgnore = this.model.get('pm_ignore');
```
martinRenou, line 63: Should we handle the `change:pm_ignore` as well?

martinRenou:
> Hey @mangecoeur, I tested an environment with PR https://github.com/jupyter-widgets/ipyleaflet/pull/1247 and Geoman was still able to modify GeoJSON layers

Does it mean we can safely merge both PRs without conflicts/collision?

python/jupyter_leaflet/src/layers/GeoJSON.ts
martinRenou: Let's handle that separately!

PR #1252, opened 2025-05-23
Default zoom level to 4 instead of 12 so that default Map() don't onl…

SylvainCorlay:
…y show sea.


SylvainCorlay:
@martinRenou

arjxn-py (approved):
Thanks, this is quite effective & works well :)

PR #1254, opened 2025-06-03
Fix CI

arjxn-py:
Continues #1248


martinRenou:
please update snapshots

PR #1255, opened 2025-06-04
CI: check comment author association instead of issue author

arjxn-py:
This PR updates the conditional logic in the `update-snapshots` GitHub Actions workflow to correctly validate the **comment author's** association with the repository, instead of the **issue/PR author's**.


martinRenou (approved):
Thanks!

